PR #10, opened 2018-12-26
Add Monix Task integration

pbylicki:
Hi,
First, really great job with this library!
I was experimenting with neotypes and monix with this straightforward implementation of Async for monix Task. Created a PR for you to see if it could be included in its current form.

Best regards,
Pawel


dimafeng:
@pbylicki that's awesome! 
Thank you!

PR #14, opened 2019-04-03
Minor changes to the Async typeclass and its instance for cats-effect

BalmungSan:
Basically I just:

+ Bump up the cats-effect and the Neo4j java driver _dependencies_.
+ Made the `AsyncExt` **implicit class** a **value class** too.
+ Remove the unnecessary `ApplicativeError[F, Throwable]` _constraint_ for `CatsAsync` instance.
+ Replaced the _"unsafe"_ typecasts on `CatsAsync.recoverWith` with a _"type-safe"_ identity map.
+ Renamed the `excpetions` package to `exceptions` - _(**Not sure if this change will break someone code, I can revert it if you want**)_.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,13 +2,13 @@ import Dependencies._
 import xerial.sbt.Sonatype._
 import ReleaseTransformations._
 
-val neo4jDriverVersion = "1.7.2"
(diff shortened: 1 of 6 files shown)


dimafeng:
@BalmungSan Thank you!

PR #16, opened 2019-04-05
Fs2 stream integration

BalmungSan:


cats-effect/src/main/scala/neotypes/cats/CatsAsync.scala
```diff
@@ -1,10 +1,9 @@
 package neotypes.cats
 
-import cats.{ApplicativeError, effect}
+import cats.effect
 import neotypes.Async
 
-class CatsAsync[F[_]](implicit F: effect.Async[F]) extends Async[F] {
-
+final class CatsAsync[F[_]](implicit F: effect.Async[F]) extends Async[F] {
```
BalmungSan: Any reason for leaving these classes public?, shouldn't they be private to at least the `neotypes` _package_?

Also, any reason for having named classes at all?, why not simply an **anonymous class** as the body of the `implicit def` ?

cats-effect/src/main/scala/neotypes/cats/CatsAsync.scala
```diff
@@ -15,7 +14,7 @@ class CatsAsync[F[_]](implicit F: effect.Async[F]) extends Async[F] {
     F.map(m)(f)
 
   override def recoverWith[T, U >: T](m: F[T])(f: PartialFunction[Throwable, F[U]]): F[U] =
-    F.recoverWith(F.map(m)(identity): F[U])(f)
+    F.recoverWith(F.map(m)(identity[U]))(f)
```
BalmungSan: IMHO, this looks nicer than the type ascription I made yesterday 😄

docs/src/main/tut/docs/streams.md
```diff
@@ -5,10 +5,12 @@ title: "Streams"
 
 # Streams
 
-Starting version `0.5.0`, you can stream the query result. As of `0.5.0`, there is only one implementation of streaming - Akka Stream.
+Starting version `0.5.0`, you can stream the query result. As of `?.?.?`, there are two implementations of streaming - [Akka Streams](https://doc.akka.io/docs/akka/current/stream/index.html) & [FS2](https://fs2.io/).
```
BalmungSan: Remember to change the version number when releasing 😉 

Also, since I do not use Akka Streams, I am not sure if that is the best link.

BalmungSan:
BTW, I believe this statement is misleading.

> **Auto-closing**: you may wish to automatically commit (or rollback on failure) the transaction once
all elements are consumed. This behaviour is provided by `stream` method available in `neotypes.Session`

Since there is not auto commit, only rollback.

cats-effect/src/main/scala/neotypes/cats/CatsAsync.scala
dimafeng: @BalmungSan good point! I haven't thought about it yet. 
I agree that access should be limited here. Would you like to submit another PR with access fixes? 😀

dimafeng:
@BalmungSan Thank you for submitting this feature implementation!

Agree with the Auto-closing node, I'll address it soon.

cats-effect/src/main/scala/neotypes/cats/CatsAsync.scala
BalmungSan: @dimafeng of course, no problem. Will do when I had the time.
However, do you want just to limit the access or going the **anonymous class** route. Which, IMHO, is better.

dimafeng: The anonymous class option sounds good to me.

PR #17, opened 2019-04-06
Cleaning code (Core)

BalmungSan:
Ok, so. I was going to refactor the **typeclasses** _instances_, to be **anonymous classes**.
But, I noticed a few _"code smells"_ (like: implicit objects, non-final case classes, too long lines, package objects, unsafe typecasts)_ that can be easily refactored. Also, there are a some places where an `Iterator` could be used to avoid intermediate collections.
And, since I am a bit free this weekend, I said "why not" - hope you like the changes.

Anyways, in the process, I realized the amount of changed lines will be big. Thus, I decided to open the PR early as a WIP, so you can be reviewing the changes as I made them if you want.
Also, If at some point. you believe there are too many for one PR, we can just merge it there and open other.


core/src/main/scala/neotypes/types.scala
```diff
@@ -1,7 +1,5 @@
 package neotypes
 
-package object types {
-
-  class Path[NODE, RELATIONSHIP](val nodes: Seq[NODE], val relationships: Seq[RELATIONSHIP], path: org.neo4j.driver.v1.types.Path)
-
+object types {
+  final class Path[NODE, RELATIONSHIP](val nodes: Seq[NODE], val relationships: Seq[RELATIONSHIP], path: org.neo4j.driver.v1.types.Path)
```
BalmungSan: Any reason for this not being a case class?

core/src/main/scala/neotypes/utils.scala
```diff
@@ -1,23 +1,32 @@
 package neotypes.utils
 
 import java.util.concurrent.CompletionException
+import java.util.function.{Consumer => JConsumer, Function => JFunction}
 
 object FunctionUtils {
```
BalmungSan: Are these utils expected to be used by users or are internal only?

core/src/main/scala/neotypes/DeferredQuery.scala
```diff
@@ -43,51 +41,52 @@ case class DeferredQuery[T](query: String, params: Map[String, Any] = Map()) {
...
-  def +(other: DeferredQueryBuilder): DeferredQueryBuilder = {
+  def +(other: DeferredQueryBuilder): DeferredQueryBuilder =
     new DeferredQueryBuilder(
-      (parts.slice(0, parts.size - 1) :+ parts.last.asInstanceOf[Query].merge(other.parts.head.asInstanceOf[Query])) ++ other.parts.tail
+      (parts.init :+ parts.last.asInstanceOf[Query].merge(other.parts.head.asInstanceOf[Query])) ++ other.parts.tail
     )
-  }
 
-  def +(other: String): DeferredQueryBuilder = {
+  def +(other: String): DeferredQueryBuilder =
```
BalmungSan: Do you believe is useful to have these concat functions?

IMHO, I would be better to remove this **Builder** and just have the `c` _interpolator_ return a **PreDeferredQuery** _(I am bad for names)_ which simply has the `query` method.
And, also create an implicit class for strings with the `query` method too.

My reasoning is that, I do not believe people will be concatenating larger queries, but instead will use a multi-line string.
And, if for some reason, they are building the query dynamically, the will be using plain strings anyways, so they may just concat the strings first and the call `query` on the result.

core/src/main/scala/neotypes/implicits.scala
```diff
@@ -0,0 +1,405 @@
...
+            Right(List.empty)
+
+          case Some(value) =>
+            sequenceAsList(
+              value
+                .values
+                .asScala
+                .iterator
+                .map { value => mapper.to("", Option(value)) }
+            )
```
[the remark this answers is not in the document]
dimafeng: Yes, I think order matters e.g.`"match (u:User) return u.firstName, u.lastName".query[List[String]].single(s)` would return firstName as a first element. 

This actually makes me think that I need to create much more unit tests to use them as a specification.

core/src/main/scala/neotypes/implicits.scala
```diff
@@ -0,0 +1,405 @@
...
+              }
+
+              val relationships = path.relationships.asScala.iterator.zipWithIndex.map {
+                case (relationship, index) => rm.to(Seq(s"relationship $index" -> new RelationshipValue(relationship)), None)
+              }
+
+              for {
+                nodes <- sequenceAsList(nodes).right
+                relationships <- sequenceAsList(relationships).right
+              } yield new types.Path(nodes, relationships, path)
```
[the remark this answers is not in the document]
dimafeng: Yes, think it's the same here - order matters

core/src/main/scala/neotypes/implicits.scala
```diff
@@ -0,0 +1,405 @@
...
+                val node = value.head._2.asNode
+                val nodes =
+                  node
+                    .keys
+                    .asScala
+                    .iterator
+                    .map(key => key -> node.get(key))
+                    .toList
+
+                (Constants.ID_FIELD_NAME -> new IntegerValue(node.id)) :: nodes
```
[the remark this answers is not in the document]
dimafeng: I believe it doesn't since these are named pairs.

core/src/main/scala/neotypes/types.scala
dimafeng: It looks like I missed `case` 😁

core/src/main/scala/neotypes/utils.scala
dimafeng: No, it's internal for `2.11` support only.

core/src/main/scala/neotypes/utils.scala
```diff
@@ -1,23 +1,32 @@
...
+    new JFunction[Throwable, Void] {
+      override def apply(e: Throwable): Void = {
+        // Execute the function.
+        e match {
+          case _: CompletionException => fn(e.getCause)
+          case _                      => fn(e)
+        }
+        // Return null, which is the only value that conforms to Void.
+        // See: https://stackoverflow.com/questions/44171262/convert-scala-unit-to-java-void/44172467#44172467
+        None.orNull
```
dimafeng: can we return `null`?

dimafeng:
@BalmungSan Thank you so much for helping with the project!

core/src/main/scala/neotypes/utils.scala
BalmungSan: If you want.

I am used to work with linters and fatal warnings. Thus, any explicit null will break the compile.
For those cases, I always use `None.orNull` which as the [scaladoc](https://www.scala-lang.org/api/current/scala/None$.html#orNull[A1%3E:A](implicitev:Null%3C:%3CA1):A1) says, such method was made to avoid an explicit use of null.

dimafeng: Makes sense. Let's leave `None.orNull`

core/src/main/scala/neotypes/DeferredQuery.scala
BalmungSan: I just finished with the new builder :D

dimafeng:
@BalmungSan looks good! Do you want me to merge the current state of this branch?

BalmungSan:
@dimafeng Let me finish with core, I am only missing `Transaction` (I believe I will have it on a couple of hours)_.
And we cam merge it.

Latter, I will open another PR with the changes on all other modules.

dimafeng:
@BalmungSan sounds good 👍

BalmungSan:
@dimafeng ok, I think I finished with `core`, unless you have any comment / suggestion.
It would be good if you could review all the changes again, just to be sure everything is ok.

Tomorrow I will open another PR for all the other modules.

core/src/main/scala/neotypes/implicits.scala
```diff
@@ -381,4 +383,46 @@ object implicits {
       new DeferredQueryBuilder(parts.toList)
     }
   }
+
+  /**
+    * Async
+    */
```
BalmungSan: I realized that it was strange that we had almost all implicits in this file, except for the ones related with `Asyc`.
Thus, I think that moving it here would be the best option, that way one only had to `import neotypes.implicits._` to use the library.

core/src/main/scala/neotypes/DeferredQuery.scala
```diff
@@ -8,9 +8,6 @@ final case class DeferredQuery[T](query: String, params: Map[String, Any] = Map.
   def list[F[_]](session: Session[F])(implicit rm: ResultMapper[T]): F[List[T]] =
     session.transact(tx => list(tx))
 
-  def seq[F[_]](session: Session[F])(implicit rm: ResultMapper[T]): F[Seq[T]] =
```
dimafeng: I thought of `seq` as a "more performant" alternative to list as it reuses underlying collection. But now I think that we're still doing lots of `map`s so there's overhead in any case.

Do you have any thought on it?

core/src/main/scala/neotypes/Transaction.scala
```diff
@@ -1,120 +1,133 @@
...
+    case s: Set[_]    => s.map(toNeoType).asJava
+    case m: Map[_, _] => m.mapValues(toNeoType).asJava
+    case o: Option[_] => o.map(toNeoType).orNull
+    case s: String    => s
+    case c: Char      => c: java.lang.Character
+    case b: Boolean   => b: java.lang.Boolean
+    case i: Int       => i: java.lang.Integer
+    case l: Long      => l: java.lang.Long
+    case f: Float     => f: java.lang.Float
+    case d: Double    => d: java.lang.Double
```
dimafeng: I think there still should be `case v: Any => v` e.g. for LocalDate, LocalTime, etc

core/src/main/scala/neotypes/implicits.scala
dimafeng: I'm thinking that this file becomes overloaded with too many different types of implementations. As far as I remember, I saw some library where all implicits were broken down into multiple classes and defined as trait members, then all of them were mixed to `object implicits`

BalmungSan: Yeah, that is what cats and most typelevel projects do.
Sounds good, but will leave it for another **PR** this one already moves too many things.

core/src/main/scala/neotypes/DeferredQuery.scala
BalmungSan: Since I am using an iterator and the resulting list is built backwards and then reversed I think it will have good performance.
Anyways, if the result is too big for it to be considered bad, a **Stream** would always be the best alternative.

Finally, I do not really like returning a `Seq`.
Is a type which you can use safely, I prefer having a more concrete type like `List` or `Set`.
For a more detailed discussion about that, see [this](https://gist.github.com/djspiewak/2ae2570c8856037a7738#generic-interfaces).

dimafeng: Sounds good

core/src/main/scala/neotypes/Transaction.scala
BalmungSan: The problem is that now, it is a **Function** from `Any` to `AnyRef`, so that won't compile.
We can do `case v: Any => v.asInstanceOf[AnyRef]` it is a _safe_ cast, since all primitive types are already handled and we could even remove them, since they will be automatically boxed. But it is still a `asInstanceOf`.

I forgot that you added some custom types, I think it would be better to add all possibilities here.
Which ones should they be, all that had an instance of **ValueMapper**?

The only problem is, does an user can create a custom type?
If so, the unsafe cast should be made at the end anyways.

dimafeng: This part is responsible for passing parameters (not result mapping) and it's currently unsafe (passed as `params: Map[String, Any]`).
I didn't come up with a good way to do it more type-safely - as I remember, **doobie** doesn't support it too. My initial idea was to use shapeless records but it's no intuitive and doesn't look concise.

I you have good idea on how to implement it I'd be happy to discuss.

BalmungSan: Yeah, yeah. It is already ugly. As I said the `asInstanceOf[AnyRef]` is _safe_ in the sense it will always be correct. So, if we can not know by had which types we could have, I think that would be the best option.

Anyways, going to sleep.
Will see your answer tomorrow, and will push the final changes ASAP.
BTW, little spoiler, I realized we are missing a `Set` **ValueMapper** and a `set` method on queries, will add them too 😉  👍

dimafeng: `asInstanceOf[AnyRef]` works for me. 
Oh yeah, seems I missed `Set`

dimafeng: @BalmungSan would you like me to make this change?

BalmungSan: @dimafeng I should have it done on a couple of hours or so.
Tomorrow on the worst case. :+1:

docs/src/main/tut/docs/types.md
```diff
@@ -16,6 +16,8 @@ title: "Supported types"
 | `scala.Option[T]                       `  | ✓              |✓                     |✓|
 | `scala.Boolean                         `  | ✓              |✓                     |✓`*`|
 | `scala.Array[Byte]                     `  | ✓              |✓                     |✓|
+| `scala.List[T: ValueMapper]            `  | ✓              |                      |✓|
+| `scala.Set[T: ValueMapper]             `  | ✓              |                      |✓|
```
BalmungSan: I am unsure if a `List[T]` or a `Set[T]` can be a **property**, and as such be a field of a **case class*.

See this: https://neo4j.com/docs/cypher-manual/current/syntax/values/

At first, they say that it is possible, for simple types. But latter they say it wont work on general _(maybe they mean that lists of lists are not valid?)_.

Maybe, you should add the ✔️ but with another _mark_ with that link?

BalmungSan:
@dimafeng Sorry for taking too long, today was complex.

Anyways, this should be ready.
I hope you could check the final changes and merge it _(if everything is ok)_ while I am asleep.
So that tomorrow I can sync #18 to merge it too.

core/src/main/scala/neotypes/Transaction.scala
dimafeng: Sorry, didn't mean to hurry you, just wanted to offer some help 👍

docs/src/main/tut/docs/types.md
dimafeng: As far as I remember, it's not possible at the moment. I expect that library users will submit a ticket if they find a way to return it 😁

dimafeng:
Great job! 👍

PR #18, opened 2019-04-08
Refactor modules

BalmungSan:
Those are very simple, just making all typeclasses instances anonymous classes and renamed all files to `implicits.scala` _(if you would like to retain the old names that can be easily done)_, plus some other minor changes.

It is funny, that the very reason why I began to made all those changes, is the last thing I had done 😅.

The idea would be to merge first the refactor to core, and then I will sync this branch with master to fix all conflicts.


dimafeng:
Awesome! Waiting on merge of #17

PR #19, opened 2019-04-15
Adding missing result mappers

BalmungSan:
Added:

+ OffsetDateTime
+ OffsetTime
+ ZonedDateTime
+ IsoDuration
+ Point

Changed the abstract classes with a two methods that returns anonymous instances of the mappers to reduce a bit of boiler plate.

In general, these are missing proper tests.
But I believe you wanted to add some tests, so I will leave it to you.

_PS: I also reordered the implicits which have a one to one relationship with methods in the Value class by alphabetical order_.

### DIFF
--- a/core/src/main/scala/neotypes/implicits.scala
+++ b/core/src/main/scala/neotypes/implicits.scala
@@ -25,10 +25,11 @@ object implicits {
     * ValueMappers
     */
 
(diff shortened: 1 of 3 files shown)


dimafeng:
Great! Thank you!

PR #21, opened 2019-04-22
Proposal to simplify calls to the stream method.

BalmungSan:
IMHO, it was very annoying to specify both, the **Stream** _(`S[_]`)_ and the **Effect** _(`F[_
`)_ type parameters, when calling the `stream` method on a query.
Especially because, always the **Stream** implied the **Effect**. Thus, it was kind of redundant.

So, somewhat inspired by the [**Partially-Applied Type**](https://typelevel.org/cats/guidelines.html) and the [**Aux**](http://gigiigig.github.io/posts/2015/09/13/aux-pattern.html) _patterns_, I come up with this proposal.

Looking forward to your comments.

### DIFF
--- a/core/src/main/scala/neotypes/implicits.scala
+++ b/core/src/main/scala/neotypes/implicits.scala
@@ -418,8 +418,10 @@ object implicits {
       new Driver[F](driver)
   }
 
-  implicit def String2QueryBuilder(query: String): DeferredQueryBuilder =
-    new DeferredQueryBuilder(List(DeferredQueryBuilder.Query(query)))
+  implicit class StringOps(val s: String) extends AnyVal {
+    def query[T]: DeferredQuery[T] =
+      DeferredQuery(query = s)
+  }
 
   implicit class CypherString(val sc: StringContext) extends AnyVal {
     def c(args: Any*): DeferredQueryBuilder = {
--- a/fs2-stream/src/main/scala/neotypes/fs2/implicits.scala
+++ b/fs2-stream/src/main/scala/neotypes/fs2/implicits.scala
@@ -1,8 +1,10 @@
 package neotypes.fs2
 
 object implicits {
-  implicit def fs2Stream[F[_]](implicit F: cats.effect.Sync[F]): neotypes.Stream[Fs2FStream[F]#T, F] =
-    new neotypes.Stream[Fs2FStream[F]#T, F] {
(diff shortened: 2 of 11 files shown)


core/src/main/scala/neotypes/implicits.scala
```diff
@@ -418,8 +418,10 @@ object implicits {
       new Driver[F](driver)
   }
 
-  implicit def String2QueryBuilder(query: String): DeferredQueryBuilder =
```
BalmungSan: I realized that having an implicit conversion was a really bad idea.
So, I just replaced both uses of it with _"safer"_ alternatives.

fs2-stream/src/main/scala/neotypes/fs2/implicits.scala
```diff
@@ -9,7 +9,7 @@ object implicits {
         fs2.Stream.repeatEval(F.suspend(value())).unNoneTerminate
 
       override def onComplete[T](s: fs2.Stream[F, T])(f: => F[Unit]): fs2.Stream[F, T] =
-        s ++ fs2.Stream.eval_(f)
+        s.onFinalize(f)
```
BalmungSan: The previous implementation was not correct. Since, in case of any error, the `f` function _(namely the rollback)_ would not be executed.

In contrast, since `onFinalize` comes from **Bracket**, it is guarantee that [_"the effectful action is executed at the end of this stream, regardless of how the stream terminates"_](https://oss.sonatype.org/service/local/repositories/releases/archive/co/fs2/fs2-core_2.12/1.0.1/fs2-core_2.12-1.0.1-javadoc.jar/!/fs2/Stream.html#onFinalize[F2[x]%3E:F[x]](f:F2[Unit])(implicitF2:cats.Applicative[F2]):fs2.Stream[F2,O]).

dimafeng:
@BalmungSan I briefly checked the code and I like it!
I'll review in more details later today. 
Thank you!

BalmungSan:
BTW, just to let you know that I also just finished an implementation of **Streaming** for `monix.reactive.Observable`, I made it over the changes made in this PR to simplify things.

Thus, as soon as I notice you merge this _(which of course should be after addressing any suggestion you had)_ I will sync my local and create a new PR with it.
So, if you can, please stay tuned for other 30 minutes or so after the merge :+1:

dimafeng:
@BalmungSan great job! I like the convince this PR brings

PR #22, opened 2019-04-23
Monix Stream

BalmungSan:
Implements the Stream typeclass for `monix.reactive.Observable`.

Disclaimer, this is my very first time using **Monix**. I basically just read the docs and searched for methods on the scaladoc that _"type-check"_.
As such, this may be a very naive implementation.

_(While the same could be said about my implementation for **FS2**, that was not my first time with it, also I use that on my daily projects)_.

### DIFF
--- a/core/src/main/scala/neotypes/DeferredQuery.scala
+++ b/core/src/main/scala/neotypes/DeferredQuery.scala
@@ -4,7 +4,7 @@ import neotypes.mappers.{ExecutionMapper, ResultMapper}
 
 import scala.collection.mutable.StringBuilder
 
-private[neotypes] final case class DeferredQuery[T](query: String, params: Map[String, Any] = Map.empty) {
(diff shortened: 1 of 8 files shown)


dimafeng:
@BalmungSan great! 👍

core/src/main/scala/neotypes/DeferredQuery.scala
```diff
@@ -4,7 +4,7 @@ import neotypes.mappers.{ExecutionMapper, ResultMapper}
 
 import scala.collection.mutable.StringBuilder
 
-private[neotypes] final case class DeferredQuery[T](query: String, params: Map[String, Any] = Map.empty) {
+final case class DeferredQuery[T] private[neotypes] (query: String, params: Map[String, Any] = Map.empty) {
```
BalmungSan: I forgot that making the class itself private, forbids creation of methods that return such instances, which is useful.
Thus, I make the class public again but made its constructor private.

The same goes for the Builder.

PR #23, opened 2019-05-06
Replacing sequence with traverse

BalmungSan:
I just remembered that `sequence(c) === traverse(c)(identity)`  and that `sequence(map(c)(f)) === traverse(c)(f)`.
And, after quick look at the code, I realized that all uses of **sequence** where made after a **map**. So, I decided to make the change.

_Anyways, since we are using Iterators, I do not expect any performance improvement for this change_.


dimafeng:
Look good 👍

PR #24, opened 2019-05-11
ResultMapper/ValueMapper derivation functions

sloshy:
Hello, just wanted to say first, thanks for the awesome project!

I was hoping to add some code that would make it a little easier to derive ResultMappers and ValueMappers from existing mappers. I took some inspiration from other libraries that have their own "Encoder/Decoder" style (mainly Circe and PureConfig) and wrote a few handy methods that should allow creating custom mappers easily.

It's not a full solution with (i.e.) Cats support, but if there's interest I would be happy to write an extra optional module for neotypes that adds some Cats instances where appropriate based on this. I hope it looks good to you, and let me know if there's anything I can tweak!

Summary of the changes:
* Added `or, map, flatMap, product, either` functions to both ResultMapper and ValueMapper.
* Changed type parameter from `T` to `A` (made some conventions a little more straightforward, particularly for Map since `T => A` looks a little weird compared to `A => B`)
* Added some basic tests for each of the new methods

If you'd like me to flesh out the tests a bit I can build on them with ScalaCheck, but I didn't want to be too disruptive with these changes.

### DIFF
--- a/core/src/main/scala/neotypes/mappers.scala
+++ b/core/src/main/scala/neotypes/mappers.scala
@@ -7,25 +7,170 @@ import scala.annotation.implicitNotFound
 import scala.reflect.ClassTag
 
 object mappers {
-  @implicitNotFound("Could not find the ResultMapper for ${T}")
-  trait ResultMapper[T] {
-    def to(value: Seq[(String, Value)], typeHint: Option[TypeHint]): Either[Throwable, T]
+  @implicitNotFound("Could not find the ResultMapper for ${A}")
+  trait ResultMapper[A] { self =>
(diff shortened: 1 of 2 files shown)


dimafeng:
@sloshy thank you!
I'll take a look soon.

sloshy:
I added some more constructors for mappers now, so it's possible to create mappers that return constant right/left values, and it's now possible to bring an implicit mapper into scope, like this:
```
val strMapper = ResultMapper[String] //implicitly loads StringResultMapper
```

dimafeng:
@sloshy looks really useful! 👍

PR #30, opened 2019-06-12
Zio integration

BalmungSan:
Implements the **Async** _typeclass_ for `zio.Task[T]` & the **Stream** _typeclass_ for `zio.stream.ZStream[Any, Throwable, T]`.

Disclaimer: Much like #22 this is my very first time using ZIO. Thus, this may be a very naive implementation, that only "type-checks".

### DIFF
--- a/monix/src/test/scala/neotypes/monix/MonixAsyncSpec.scala
+++ b/monix/src/test/scala/neotypes/monix/MonixAsyncSpec.scala
@@ -6,27 +6,25 @@ import neotypes.monix.implicits._
 import neotypes.implicits._
 import neotypes.BaseIntegrationSpec
 import org.neo4j.driver.v1.exceptions.ClientException
-import org.scalatest.FlatSpec
+import org.scalatest.AsyncFlatSpec
 
-import scala.concurrent.duration._
-
-class MonixAsyncSpec extends FlatSpec with BaseIntegrationSpec {
+class MonixAsyncSpec extends AsyncFlatSpec with BaseIntegrationSpec {
   it should "work with Task" in {
     val s = driver.session().asScala[Task]
 
-    val string =
-      "match (p:Person {name: 'Charlize Theron'}) return p.name"
-        .query[String]
-        .single(s)
-        .runSyncUnsafe(5 seconds)
+    "match (p:Person {name: 'Charlize Theron'}) return p.name"
+      .query[String]
+      .single(s)
+      .runToFuture
+      .map {
+        name => assert(name == "Charlize Theron")
+      }
 
-    assert(string == "Charlize Theron")
-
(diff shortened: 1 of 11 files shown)


monix/src/test/scala/neotypes/monix/MonixAsyncSpec.scala
```diff
@@ -6,27 +6,25 @@ import neotypes.monix.implicits._
 import neotypes.implicits._
 import neotypes.BaseIntegrationSpec
 import org.neo4j.driver.v1.exceptions.ClientException
-import org.scalatest.FlatSpec
+import org.scalatest.AsyncFlatSpec
 
-import scala.concurrent.duration._
-
-class MonixAsyncSpec extends FlatSpec with BaseIntegrationSpec {
+class MonixAsyncSpec extends AsyncFlatSpec with BaseIntegrationSpec {
```
BalmungSan: Since we have been using the `AsyncFlatSpec` by turning all effects into `Futures` in everywhere else. I tough it would be better to leave that as an _"standard"_. Thus, I refactored this one that was the only one which did not use it.

zio/src/main/scala/neotypes/zio/implicits.scala
```diff
@@ -0,0 +1,30 @@
...
+        m.map(f)
+
+      override def recoverWith[T, U >: T](m: Task[T])(f: PartialFunction[Throwable, Task[U]]): Task[U] =
+        m.catchSome(f)
+
+      override def failed[T](e: Throwable): Task[T] =
+        Task.die(e)
+
+      override def success[T](t: => T): Task[T] =
+        Task(t)
```
ghostdogpr: If t can’t fail, you can use `Task.succeedLazy`

zio/src/main/scala/neotypes/zio/implicits.scala
```diff
@@ -0,0 +1,30 @@
...
+        m.flatMap(f)
+
+      override def map[T, U](m: Task[T])(f: T => U): Task[U] =
+        m.map(f)
+
+      override def recoverWith[T, U >: T](m: Task[T])(f: PartialFunction[Throwable, Task[U]]): Task[U] =
+        m.catchSome(f)
+
+      override def failed[T](e: Throwable): Task[T] =
+        Task.die(e)
```
ghostdogpr: I think the ZIO equivalent here would be `Task.fail`. `die` is for defects and will interrupt the whole fiber.

BalmungSan:
Great, thanks @ghostdogpr I will make the changes in a couple of hours.

dimafeng:
@BalmungSan awesome!
@ghostdogpr thanks!

PR #32, opened 2019-06-12
refined integration

### LINKED ISSUE #31: Feature: refined support
I would be great if this could be integrated _(probably with an external module)_ with [**refined**](https://github.com/fthomas/refined).

I see two big opportunities.
1. Using refined types in query builders. For example:

```scala
type Level = Int Refined Interval.Closed[W.`1`.T, W.`99`.T]
val L1: Level = 1

// This does not work.
c"CREATE (level: Level { value: ${L1} })"

// Right now, we have to do this to manually unwrap the Int from the Refined,
// which for many values is kind of tedious.
c"CREATE (level: Level { value: ${L1.value} })"
```

2. Returning refined values from queries, if the value does not satisfy the refinement return a Throwable.

```scala
type Level = Int Refined Interval.Closed[W.`1`.T, W.`99`.T]

c"MATCH (level: Level) RETURN level.value".query[Level].list(session)
// res: Future[List[Level]]
```

BalmungSan:
Fixes #31 

Ok, this seems to work.
But, there is still work to do, first I do not think you like the idea that it is tightly coupled in `core`.
Also, this needs more tests, like a refined inside a case class and validating failed cases.

But, I am opening this for early feedback.

### DIFF
--- a/core/src/main/scala/neotypes/Transaction.scala
+++ b/core/src/main/scala/neotypes/Transaction.scala
@@ -3,6 +3,7 @@ package neotypes
 import java.util.{Map => JMap}
 import java.util.concurrent.CompletionStage
 
+import eu.timepit.refined.api.Refined
 import neotypes.mappers.{ExecutionMapper, ResultMapper}
 import neotypes.utils.traverse.{traverseAsList, traverseAsSet}
 import neotypes.utils.stage._
@@ -130,16 +131,17 @@ object Transaction {
     params.mapValues(toNeoType).asJava
 
   private def toNeoType(value: Any): AnyRef = value match {
-    case s: Seq[_]    => s.map(toNeoType).asJava
-    case s: Set[_]    => s.map(toNeoType).asJava
-    case m: Map[_, _] => m.mapValues(toNeoType).asJava
-    case o: Option[_] => o.map(toNeoType).orNull
+    case s: Seq[_]        => s.map(toNeoType).asJava
+    case s: Set[_]        => s.map(toNeoType).asJava
+    case m: Map[_, _]     => m.mapValues(toNeoType).asJava
+    case o: Option[_]     => o.map(toNeoType).orNull
+    case r: Refined[_, _] => toNeoType(r.value)
     // The following type cast is sound,
     // since AnyRef is equivalent to java.lang.Object
     // and all values coming from a class can be upcasted to it.
     // For value types (Subtypes of AnyVal),
     // this cast ends up boxing them to their objectual representation.
     // A such, this will never fail in runtime.
-    case v            => v.asInstanceOf[AnyRef]
+    case v                => v.asInstanceOf[AnyRef]
   }
 }
--- a/core/src/main/scala/neotypes/Transaction.scala
+++ b/core/src/main/scala/neotypes/Transaction.scala
@@ -4,7 +4,7 @@ import java.util.{Map => JMap}
 import java.util.concurrent.CompletionStage
 
 import mappers.{ExecutionMapper, ResultMapper}
-import types.NeoType
+import types.QueryParam
 import utils.traverse.{traverseAsList, traverseAsSet}
 import utils.stage._
 
@@ -17,7 +17,7 @@ import scala.collection.JavaConverters._
 final class Transaction[F[_]](transaction: NTransaction)(implicit F: Async[F]) {
   import Transaction.{convertParams, recordToSeq}
 
-  def execute[T](query: String, params: Map[String, NeoType] = Map.empty)
+  def execute[T](query: String, params: Map[String, QueryParam] = Map.empty)
                 (implicit executionMapper: ExecutionMapper[T]): F[T] =
     F.async { cb =>
       transaction
@@ -27,7 +27,7 @@ final class Transaction[F[_]](transaction: NTransaction)(implicit F: Async[F]) {
         .recover { ex: Throwable => cb(Left(ex)) }
     }
 
-  def list[T](query: String, params: Map[String, NeoType] = Map.empty)
+  def list[T](query: String, params: Map[String, QueryParam] = Map.empty)
              (implicit resultMapper: ResultMapper[T]): F[List[T]] =
     F.async { cb =>
       transaction
@@ -42,7 +42,7 @@ final class Transaction[F[_]](transaction: NTransaction)(implicit F: Async[F]) {
         }.recover { ex: Throwable => cb(Left(ex)) }
     }
 
-  def set[T](query: String, params: Map[String, NeoType] = Map.empty)
+  def set[T](query: String, params: Map[String, QueryParam] = Map.empty)
             (implicit resultMapper: ResultMapper[T]): F[Set[T]] =
     F.async { cb =>
       transaction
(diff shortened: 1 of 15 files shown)


core/src/main/scala/neotypes/Transaction.scala
```diff
@@ -130,16 +131,17 @@ object Transaction {
...
   private def toNeoType(value: Any): AnyRef = value match {
-    case s: Seq[_]    => s.map(toNeoType).asJava
-    case s: Set[_]    => s.map(toNeoType).asJava
-    case m: Map[_, _] => m.mapValues(toNeoType).asJava
-    case o: Option[_] => o.map(toNeoType).orNull
+    case s: Seq[_]        => s.map(toNeoType).asJava
+    case s: Set[_]        => s.map(toNeoType).asJava
+    case m: Map[_, _]     => m.mapValues(toNeoType).asJava
+    case o: Option[_]     => o.map(toNeoType).orNull
+    case r: Refined[_, _] => toNeoType(r.value)
```
BalmungSan: I see a big problem here in order to make the refined integration on an module outside `core`. The mappers can be easily moved to another module. But, this part here not.

Any ideas?

BalmungSan:
Ok so, I think that for tests and development I am done.

The only thing that is missing is moving it into its own module...
But, I still do not have any ideas of how to do that for the `toNeoType` mapping function.
What about you @dimafeng ?

dimafeng:
@BalmungSan apologies for delay with this PR. I have some thought re `toNeoType`, promise to get back to this by tomorrow night

core/src/main/scala/neotypes/Transaction.scala
dimafeng: I fully agree with this being moved to a separate module.
I'm thinking about these two approachs:

**Option 1** 
1. Add a partial function or a sequence of partial functions to `Transaction` as a constructor parameter. So that the `toNeoType` would check if there are anything was injected for overwriting types.
2. Injection part - since the transaction initialization can be done on any level (driver, session, transaction), it seems that that parameter should be passed down to tx from any top level.
I'm currently inclined toward creating an extension method `withRefined` that "adds proper converters to the scope"
```scala
import neotypes.refined.implicits // containes mappers + `withRefined`
...
driver.asScala[Future].withRefined
...
// or
...
driver.session().asScala[Future].withRefined
```
**Option 2 (I don't know if it's possible)** 
1. Add an additional typeclass :`final class Transaction[F[_]](transaction: NTransaction)(implicit F: Async[F], converter: converter: ToNeoTypes[F]) {`
2. (*I'm not completely sure re this, I'm not an expert but I saw something like this but never thought how it's implemented 😁*) Create a DSL that combines typeclasses to have it like this `driver.asScala[Future and RefinedSupport]`

@BalmungSan what do you think? 😁

BalmungSan:
@dimafeng If we are going to refactor it, I think it would be worth exploring the **Typeclass** alternative.
That way it would be more extensible. For example, I already see use cases for that, like adding support to **Cats** `NonEmptyList`, `Chain` & `EmptyChain`. And anyone could provide its own type conversions.

Also, I think it should be somewhat related to the **Mappers**, like if there is a way to retrieve some value from Neo4j as a Scala type. Then, there should be a way to turn that value into a Neo4j.
That would open possibilities like doing this:

```
final case class User(name: String)

c"CREATE ${User("Luis")}".query[Unit].execute(s)
```

But that seems extremely too board for now.

For now, I will play a bit with your proposal.
However, this weekend is fathers day, thus it could take me a while.

dimafeng:
@BalmungSan sounds good 👍

BalmungSan:
Ok so, I end up adding a new `ParameterMapper` **typeclass** that is in charge of casting _Scala_ values into valid _Neo4j_ parameters & the new `NeoType` wrapper class that represent them.
To ensure compile-time type safety, there are no public ways of creating instances of a `NeoType` without a `ParameterMapper` for the input type. Also, there aren't any ways to create more `ParameterMappers` except for the `contramap` method _(which is the correct way for an user to support new types)_ & the `constant` method, which maybe could be useful for testing... _(I still not sure if it should be removed too, what do you think)_.

Additionally, I rewrote the `CypherStringInterpolator` to use a **Macro**.  
Which will ensure that the written query is type-safe. 😎 

I also added the instances for common types. It is still pending to check if there could be one for case classes and how that could work, I will leave for latter.
BTW, the **implicits.scala** file is now extremely long. I will split it into a couple of traits, but that will be done in another PR after this gets merged.

Finally, I am still missing the splitting of the refined support into its own module _(but that that should be pretty easy)_, cleaning the code and tune the documentation.
I am hopping to do that tomorrow afternoon.

Anyways, it would be great if you can review the changes and leave feedback meanwhile.
@dimafeng

dimafeng:
@BalmungSan wow, you've put a lot of work into this! I've briefly reviewed and realized that there's no way to understand changes without loading it into an IDE and play with the code.
Overall, I agree with the direction you picked - more type safety is better!

BalmungSan:
Ok, unless you have any feedback, I think I am done here 🚀 
This became longer that I would expect 😅 but I hope you like the changes.

I am still missing the documentation, I believe I will be pushing it tomorrow.
_(Anyways, if you want to merge this before the documentation no problems, I will just create a new PR for it)_.

dimafeng:
@BalmungSan looks good to me! And I like the idea with the macro 👍 

Got some minor comments (not even sure it needs to be discussed within this PR): 

1) Have one thought on `NeoType` - `params: Map[String, NeoType]` is part of public interface of `Transaction`, curious if this needs somehow be simplified (e.g. via implicit conversions which I don't like) or maybe even more aggressively, all `Transaction`s public methods should be hidden from public API (so people would only use `DeferredQuery`)? I'd appreciate your thoughts on it. Meanwhile I'm merging this great PR*
2) Naming of `NeoType` - would you like to change it to any of `TypedParameter`/`QueryParameter`/`TypedParam`/`QueryParam`, just thought it would be a bit more clearer when I first read it.

BalmungSan:
> Naming of NeoType - would you like to change it to any of TypedParameter/QueryParameter/TypedParam/QueryParam, just thought it would be a bit more clearer when I first read it.

Being honest, I just name it `NeoType` because the original method was named `toNeoType`, but I forgot I should have renamed it 😅 - Personally, I like `QueryParam`.

However...

> Have one thought on NeoType - params: Map[String, NeoType] is part of public interface of Transaction, curious if this needs somehow be simplified (e.g. via implicit conversions which I don't like) or maybe even more aggressively, all Transactions public methods should be hidden from public API (so people would only use DeferredQuery)?

I only added the `NeoType` at all, just because those methods are public. In my first draft, the `ParameterMapper` would return an `AnyRef` directly, but then I realized that those parameters could be passed by other means.

IMHO, removing all that would simplify the API and the mappers a lot _(you can see that many mappers that depend on others, like the one for `List`, have to wrap and unwrap the internal values)_. But, then you would need to remove the `withTransaction` method on the `Session` class too.
Personally, I only use the `DeferredQueryBuilder`, but I am not sure if those methods are useful for other people.

I believe that is a really strong decision, which you as the author should take carefully. All that I can say is that, I do not use it, and remove it would simplify the code.
Also, I agree that this should be addressed on another PR. Thus, for now I will just do the rename and add a `sealed`  that I missed for the `ParameterMapper` **trait**.

PR #33, opened 2019-06-20
Implicits refactor

BalmungSan:
I tried to be as granular as possible, So, people who like to have the minimum possible in the scope _(like me)_, can selectively add import by import. And for people who like it batteries included, there are a couple of `all` objects that will bring all in at once.
Also, I tried to follow conventions from other projects like **cats**.

I think I gave it proper structure and names to each import, but naming has always been my _Achilles heel_ 😅
Thus, I am open to any suggestion you may have.

**PS:** Remember to release before merging this. But after that, if you are fine with the changes, feel free to sync this with master by yourself and merging it right away. 👍

@dimafeng


dimafeng:
@BalmungSan sorry for the delay with this one (it's been a really busy week). I'll try to finish review by tomorrow. 

Btw, the new version was released yesterday.

dimafeng:
@BalmungSan awesome! Great job!

sloshy:
Hey this PR is closed, I know, but just wanted to say thanks @BalmungSan for this PR because it fixed a really strange implicit conflict bug I was experiencing. I had a function `def flatMapF[F[_]: cats.effect.Async](f: F[Any]) = f.flatMap(_ => cats.effect.Async[F].unit)` and it couldn't figure out which `flatMap` to call due to the neotypes implicit imports and got confused. Now with the implicits refactor everything works the way I expect and my code compiles and/or doesn't give IntelliJ confusion anymore. Thanks again!

sloshy:
Actually false alarm - bug still exists, but the import changes do make it easier to avoid it by a fair margin. Added as issue #35 for now

PR #34, opened 2019-06-25
Vector support, Map query result & cats.data instances

BalmungSan:
Ok, so this is another **big** PR...
_(I am sorry, I know those are hard to review, but it was too convenient to do all these at once.)_

This PR include 5 changes _(each per commit)_:

+ Adding support for `Vector`:
  - As a field of a case class _(`ValueMapper`)_.
  - As a query result _(`ResultMapper`)_.
  - As a parameter _(`ParameterMapper`)_.
  - As a query execution _(method in `Transaction` & `DeferredQuery`)_.

+ Adding support for `Map` as a query execution _(method in `Transaction` & `DeferredQuery`)_.
   Only if `T <:< (K, V)`.

+ Reducing a little bit of boilerplate in the tests.

+ Moving the **cats-effect** implicits from `neotypes.cats.implicits._` to `neotypes.cats.effect.implicits._`.
    _(This is another breaking change!!)_

+ Adding a new module **neotypes-cats-data** which provides support for various types inside the `cats.data` package.

  -  `Chain`
  - `Const`
  - `NonEmptyChain`
  - `NonEmptyList`
  - `NonEmptyMap`
  - `NonEmptySet`
  - `NonEmptyVector`

@dimafeng 

> **PS:** As I already said, each commit should have been in its own PR. Thus, I believe it would be easier to review if instead of looking at all changes at once, you check commit by commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ For early adopters:
 |`"com.dimafeng" %% "neotypes-monix-stream" % version`|result streaming for Monix Observables.|
 |`"com.dimafeng" %% "neotypes-zio-stream" % version`|result streaming for ZIO ZStreams.|
 |`"com.dimafeng" %% "neotypes-refined" % version`|support to insert and retrieve refined values.|
+|`"com.dimafeng" %% "neotypes-cats-data" % version`|support to insert and retrieve `cats.data` values.|
 
 **Scala lightweight, type-safe, asynchronous driver (not opinionated on side-effect implementation) for neo4j**.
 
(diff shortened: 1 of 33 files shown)


dimafeng:
@BalmungSan sorry, I'm still reviewing... 😞 Don't have experience with `cats.data` but it's good chance to learn it 👍 

I reran the tests and now they're green! Great job!

BalmungSan:
@dimafeng no worries, I know it is a big one. Take your time :smile:

> Don't have experience with `cats.data` but it's good chance to learn it.

Cool :sunglasses:

dimafeng:
@BalmungSan awesome. looks great!

Issue #35, opened 2019-06-26
neotypes.Async implicit imports cause divergent implicits


sloshy:
I have code that uses `cats-effect` with higher-kinded types in most of my functions. Here's an example that should compile fine using the `cats.effect.Async` typeclass:

```scala
import cats.implicits._
import cats.effect._

def flatMapF[F[_]: Async](f: F[Any]): F[Unit] = f.flatMap(_ => Async[F].unit)
```

This code compiles and works fine unless I import the implicits for `neotypes.Async` by importing `neotypes.implicits.all._` or `neotypes.implicits.syntax.all._`. The code fails to compile due to a divergent implicit leading the compiler to not know which `flatMap` method to call.

Currently using the latest snapshot build from master, built locally (`0.9.1-SNAPSHOT`).

sloshy:
A simple workaround for the time being - to me and anyone else who runs into this - is to only import the implicits you need in a specific scope instead of the whole suite of implicits at once everywhere, such as only `ResultMapper`s or string syntax.

BalmungSan:
So the problem is as follows:

When you call `flatMap` over `fa`, since it does not have the `flatMap` method, the compiler will try to find an _implicit conversion_ from it to something that does has such method.
Then it will find both `cats.syntax.FlatMapSyntax.catsSyntaxFlatMapOps` & `neotypes.implicits.syntax.AsyncSyntax.neotypesSyntaxAsync`, and since your `F[_]` has a context bound of `cats.effect.Async[F]` then both are applicable.

Being honest, I do not think there could be other option except for do not importing the **neotypes** _implicit_. Specially because, in this case, they are equivalent.
The implicits refactor _(#33 )_ allowed to import everything you need and exclude this one. Before it, that would be impossible in user code.
I know that could be tedious, but it is the only solution, excluding it from the implicit scope.

The only other _"alternative"_ would be to have some kind of `allButAsync` object, or to move the `AsyncSyntax` trait outside of the `syntax` package. However, that is essentially the same as above, but more ergonomic. 
Also, IMHO, these feel strange, and solve a very specific issue, that maybe it would be better to leave it to the users instead.

For allowing that, we could remove the `private[neotypes]` from all traits, so users could do this:

```scala
package my.app.implicits

import neotypes.implicits.mappers.AllMappers
import neotypes.implicits.syntax.{CypherSyntax, DriverSyntax, StringSyntax}

[...]

PR #37, opened 2019-06-28
Low priority trick + cats implicits test

### LINKED ISSUE #36: Divergent implicits when using `cats.implicits._`
I've read #35 and I understand it refers to the `import neotypes.Async._`? I am currently experiencing the same issue but also with `import neotypes.implicits._` using version `0.9.0`.

Here's the original code I had using the Neo4j java driver:

```scala
import cats.Applicative
import cats.effect._
import cats.effect.implicits._
import cats.implicits._
import org.neo4j.driver.v1._

object Neo {

  case class NeoConfig(uri: String, user: String, pass: String)

  // TODO: do proper logging
  def putStrLn[F[_]: Sync, A](a: A): F[Unit] = Sync[F].delay(println(a))

  private def loadConfig[F[_]: Applicative]: F[NeoConfig] =
    NeoConfig("bolt://localhost:7687", "neo4j", "test").pure[F]

  def mkClient[F[_]: Sync]: Resource[F, NeoClient[F]] =
    Resource.make[F, NeoClient[F]](
      loadConfig[F].flatMap { c =>
        putStrLn("NeoClient[F] >> Connecting to Neo4j...") *>
          Sync[F]
            .delay(GraphDatabase.driver(c.uri, AuthTokens.basic(c.user, c.pass)))
            .map(new LiveNeoClient(_))
            .widen[NeoClient[F]] <*
          putStrLn("NeoClient[F] >> Connected")
      }
    )(_.close)

}

trait NeoClient[F[_]] {
  def close: F[Unit]
  def transaction: Resource[F, Transaction]
}

class LiveNeoClient[F[_]: Sync] private[persistence] (driver: Driver) extends NeoClient[F] {
  def close: F[Unit] = Neo.putStrLn("NeoClient[F] >> Closing connection") *> Sync[F].delay(driver.close())
  def transaction: Resource[F, Transaction] =
    Resource.make(Sync[F].delay(driver.session.beginTransaction))(tx => Sync[F].delay(tx.close))
}
```

This compiles fine until I add any of the imports `neotypes.Async._` and/or `neotypes.implicits._`

```
[error] /workspace/.../Neo.scala:22:17: type mismatch;
[error]  found   : F[Neo.NeoConfig]
[error]  required: ?{def flatMap: ?}
[error] Note that implicit conversions are not applicable because they are ambiguous:
[error]  both method toFlatMapOps in trait ToFlatMapOps of type [F[_], C](target: F[C])(implicit tc: cats.FlatMap[F])cats.FlatMap.Ops[F,C]{type TypeClassType = cats.FlatMap[F]}
[error]  and method AsyncExt in object implicits of type [F[_], T](m: F[T])neotypes.implicits.AsyncExt[F,T]
[error]  are possible conversion functions from F[Neo.NeoConfig] to ?{def flatMap: ?}
[error]       loadConfig[F].flatMap { c =>
[error]                 ^
[error] one error found
[error] (Compile / compileIncremental) Compilation failed
[error] Total time: 2 s, completed Jun 28, 2019 10:58:31 AM
```

Are there any known workarounds to this? I see there's been some work done in #33 but the error still seems to be there.

gvolpe:
I can confirm #36 is fixed in `master`, not in `0.9.0`. Anyway, I thought a test should be added to validate this is not broken in future changes. 

I also changed the way the instances are declared using what's known as the *low priority trait / abstract class*. Ideally all the other instances should be declared in this way (same as other FP libraries get around divergent implicits while remaining binary compat).

Closes #36

### DIFF
--- a/cats-effect/src/main/scala/neotypes/cats/implicits.scala
+++ b/cats-effect/src/main/scala/neotypes/cats/implicits.scala
@@ -1,7 +1,9 @@
 package neotypes.cats
 
-object implicits {
-  implicit def catsAsync[F[_]](implicit F: cats.effect.Async[F]): neotypes.Async[F] =
+object implicits extends CatsInstances
+
+private[cats] abstract class CatsInstances {
+ implicit def catsAsync[F[_]](implicit F: cats.effect.Async[F]): neotypes.Async[F] =
     new neotypes.Async[F] {
       override def async[T](cb: (Either[Throwable, T] => Unit) => Unit): F[T] =
         F.async(cb)
--- a/cats-effect/src/test/scala/neotypes/cats/CatsImplicitsSpec.scala
+++ b/cats-effect/src/test/scala/neotypes/cats/CatsImplicitsSpec.scala
@@ -1,0 +1,44 @@
+package neotypes.cats
+
+import cats.Applicative
+import cats.effect._
+import cats.effect.implicits._
+import cats.implicits._
+import neotypes.cats.implicits._
+import neotypes.Async._
+import neotypes.implicits._
+import neotypes.implicits.all._
+import neotypes.{BaseIntegrationSpec, Session}
+import org.neo4j.driver.v1.exceptions.ClientException
+import org.scalatest.AsyncFlatSpec
+import cats.Monad
+
+class CatsImplicitsSpec extends AsyncFlatSpec with BaseIntegrationSpec {
+  it should "work with cats implicits and neotypes implicits" in {
+    def test1[F[_]: Applicative]: F[Unit] = Applicative[F].unit
+
+    def test2[F[_]: Monad]: F[Unit] = ().pure[F]
+
+    def makeSession[F[_]: Async]: F[Session[F]] =
+      (test1[F] *> test2[F]).as(driver.session().asScala[F])
+
+    val s = makeSession[IO].unsafeRunSync()
+
+    """match (p:Person {name: "Charlize Theron"}) return p.name"""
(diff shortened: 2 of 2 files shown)


gvolpe:
I made the test fail by adding a call to `flatMap` with the mix of imports as demonstrated in #35 . This should pass once the issue is fixed.

cats-effect/src/main/scala/neotypes/cats/implicits.scala
```diff
@@ -1,7 +1,9 @@
 package neotypes.cats
 
-object implicits {
-  implicit def catsAsync[F[_]](implicit F: cats.effect.Async[F]): neotypes.Async[F] =
+object implicits extends CatsInstances
+
+private[cats] abstract class CatsInstances {
```
BalmungSan: Yes I was going to do this in the next PR.
See the end of this [comment](https://github.com/neotypes/neotypes/issues/35#issuecomment-506402695)

> Also, I just noticed that, the implicits of each module should also be splitted in both a trait and an object extending such trait.

gvolpe:
Go for it @BalmungSan ! I just published a local version to keep the work going.

dimafeng:
@gvolpe thank you for the contributions! I glad to see that you guys came to an agreement

@BalmungSan would you like me to close this and wait for your version?

BalmungSan:
@dimafeng yes, I should be working on that the day after tomorrow.
If you could take a look to #38 before that, it would be great.
I would like to include that too.

Issue #38, opened 2019-06-28
Creation of session should return `F[Session[F]]`


gvolpe:
Creation of mutable state should be wrapped in `F` for the reasons stated [here](https://typelevel.org/blog/2018/06/07/shared-state-in-fp.html) and [here](https://www.reddit.com/r/scala/comments/be6qxp/stm_software_transactional_memory_examples_in/). Fabio Labella also gave a [great talk](https://vimeo.com/294736344) on the topic.

```scala
val session: Session[IO] = driver.session().asScala[IO] // this can leak

val makeSession: IO[Session[IO]] = driver.session.asScala[IO] // this is the action of creating a session

makeSession.flatMap { s => // here we create a region of sharinn
  useSession(s) // whoever needs access to the session should take it as a parameter 
 // in this way the sharing of the session is under our control in a narrow scope (inside `flatMap`)
}
```

Right now I get around by just wrapping its creation in `IO` but ideally the library should do this.

This probably doesn't make sense when using `Future` since it breaks referential transparency but for the rest of us using FP libraries this is important.

Let me know if I can be of any help.
```

BalmungSan:
Could you please expand why it should be wrapped in the effect itself?
I was planning in adding a `session` method to the **Driver**, to create a **Session** directly.
Thus, that could be a good place to add that.

However, I still do not understand, what is shared about the session?
I believe it is kind of immutable. However, I really do not know how the internal Neo4j session is implemented just the wrapper in this library.

gvolpe:
@BalmungSan have you read the articles I shared or just watched the talk I linked? It's all there :)

BalmungSan:
@gvolpe well yes, but still I do not see the reason for the library to do that.
Usually, I will have a class that encapsulates all **Neo4j** operations, I will just pass the unique instance to my services as the `HttpClient` example.
In that class I will receive the `Session` object in the constructor, and I will make that constructor private.
And the companion object will have a factory returning the instance of my class wrapped on a `Resource[F, C]` which will close the session after use.

gvolpe:
> well yes, but still I do not see the reason for the library to do that.

Creation of mutation *is a side-effect*. This should be supported by the library unless having mutability and side-effects is what we want to promote.

What you're doing is fine and is what I have to do as well. But I don't agree this should be let off to the user.

---

[...]

PR #39, opened 2019-07-02
Refactoring the public API

### LINKED ISSUE #36: Divergent implicits when using `cats.implicits._`

BalmungSan:
Well, here comes another BIG PR... 😞 - _I promise this is (one of) the last_ 😅

This PR includes:

- [x] Removal of the `async` **syntax** from the public API.
- [x] Making all **implicits** traits public.
- [x] Refactoring the  way to create instances of `Driver[F]` & `Session[F]` _(more on this down)_.
- [x] Adding a test case to ensure the neotypes **implicits** do not collide with the ones from cats.
- [x] Splitting the modules **implicits** into a public trait and a default object.

I think code speaks better than me, here is an extract from the updated documentation.

### For future

```scala
import neotypes.implicits.mappers.results._ // Brings the implicit ResultMapper[String] instance into the scope.
import neotypes.implicits.syntax.string._ // Provides the query[T] extension method.
import scala.concurrent.Future
import scala.concurrent.ExecutionContext.Implicits.global

val program: Future[String] = for {
  driver <- GraphDatabase.driver[Future]("bolt://localhost:7687", AuthTokens.basic("neo4j", "****"))
  session <- driver.session()
  data <- "match (p:Person {name: 'Charlize Theron'}) return p.name".query[String].single(session)
  _ <- session.close()
  _ <- driver.close()
} yield data

val data: String = Await.result(program, 1 second)
```

### For IO

```scala
import cats.effect.{Resource, IO}
import neotypes.Session
import neotypes.cats.effect.implicits._ // Brings the implicit neotypes.Async[IO] instance into the scope.
import neotypes.implicits.mappers.results._ // Brings the implicit ResultMapper[String] instance into the scope.
import neotypes.implicits.syntax.string._ // Provides the query[T] extension method.

val session: Resource[IO, Session[IO]] = for {
  driver <- Resource.make(
    GraphDatabase.driver[IO]("bolt://localhost:7687", AuthTokens.basic("neo4j", "****"))
  ) { driver => driver.close() }

  session <- Resource.make(
    driver.session()
  ) { session => session.close() }
} yield session

val program: IO[String] = session.use { s =>
  "match (p:Person {name: 'Charlize Theron'}) return p.name".query[String].single(s)
}

val data: String = program.unsafeRunSync()
```

@dimafeng What do you think?
_(Oh, I forgot to mention that I had to hack the tests a little to make them work with the new API)_.

-----------------------

Closes #36 & #38 
Supersedes #37 

Thanks again @gvolpe for the test case and the discussion about wrapping the creation of **Session** / **Driver** inside an `F`.
I would really appreciate if you could take the time to review the changes.


docs/src/main/tut/docs.md
```diff
@@ -13,33 +13,41 @@ title: "Documentation"
 
 ## Session creation
 
-**neotypes** adds an extension method (`.asScala[F[_]: Async]`) to `org.neo4j.driver.v1.Session` and `org.neo4j.driver.v1.Driver` that allows to build a `neotypes`'s session and driver wrappers.
-You can parametrize `asScala` by any type that you have a typeclass `neotypes.Async` implementation for.
-The typeclass implementation for `scala.concurrent.Future` is built-in _(please read more about [side effect implementations](docs/alternative_effects.html))_.
+**neotypes** provides the `GraphDatabase` **factory** for **Neo4j** `Session` and `Driver` wrappers.
+You can create wrappers for any type `F[_]` for which you have an implementation of the `neotypes.Async` **typeclass**.
+The implementation for `scala.concurrent.Future` is built-in _(for other types, please read [side effects](docs/alternative_effects.html))_.
```
gvolpe: I'd replace "side effects" for just "effects" or "alternative effects".

docs/src/main/tut/docs/alternative_effects.md
```diff
@@ -11,62 +11,105 @@ title: "Side effects"
...
-import neotypes.implicits.syntax.session._ // Provides the asScala[F[_]] extension method.
 import neotypes.implicits.syntax.string._ // Provides the query[T] extension method.
 import scala.concurrent.Future
+import scala.concurrent.ExecutionContext.Implicits.global
 
-val s = driver.session().asScala[Future]
+val program: Future[String] = for {
+  driver <- GraphDatabase.driver[Future]("bolt://localhost:7687", AuthTokens.basic("neo4j", "****"))
+  session <- driver.session()
+  data <- "match (p:Person {name: 'Charlize Theron'}) return p.name".query[String].single(session)
```
gvolpe: Not that I care too much about `Future` :smile: but if this is the official example shouldn't it be shown that the session and driver are properly closed? In this example, if there's an error in the query then there's a session / driver leak.

docs/src/main/tut/docs/alternative_effects.md
```diff
@@ -11,62 +11,105 @@ title: "Side effects"
...
+import neotypes.cats.effect.implicits._ // Brings the implicit neotypes.Async[IO] instance into the scope.
 import neotypes.implicits.mappers.results._ // Brings the implicit ResultMapper[String] instance into the scope.
-import neotypes.implicits.syntax.session._ // Provides the asScala[F[_]] extension method.
 import neotypes.implicits.syntax.string._ // Provides the query[T] extension method.
 
-val s = driver.session().asScala[IO]
+val session: Resource[IO, Session[IO]] = for {
+  driver <- Resource.make(
+    GraphDatabase.driver[IO]("bolt://localhost:7687", AuthTokens.basic("neo4j", "****"))
+  ) { driver => driver.close() }
```
gvolpe: I'd probably simplify it as it's always kind of the same `(_.close)`

docs/src/main/tut/docs/alternative_effects.md
```diff
@@ -11,62 +11,105 @@ title: "Side effects"
...
+
+  session <- Resource.make(
+    driver.session()
+  ) { session => session.close() }
```
gvolpe: Ditto `(_.close)`

gvolpe: There are a few more in other files as well.

docs/src/main/tut/docs/alternative_effects.md
gvolpe: And maybe `flatMap` is more idiomatic here as you only require one (plus there's a variable shadowing):

```scala
val session: Resource[IO, Session[IO]] = 
  Resource.make(
    GraphDatabase.driver[IO]("bolt://localhost:7687", AuthTokens.basic("neo4j", "****"))
  )(_.close).flatMap { driver =>
    Resource.make(driver.session)(_.close)
  }
```

docs/src/main/tut/docs.md
BalmungSan: I think I prefer _"alternative effects"_. However, since this is documentation I will let the final vow to @dimafeng

docs/src/main/tut/docs/alternative_effects.md
BalmungSan: Uhm... you are right, I really haven't used `Future` in my life, I am just used to the safety of `IO` 😅 

Would this be a better example?

```scala
val program = for {
  driver <- GraphDatabase.driver[Future]("bolt://localhost:7687", AuthTokens.basic("neo4j", "****"))
  result <- driver.readSession { session =>
    "match (p:Person {name: 'Charlize Theron'}) return p.name".query[String].single(session)
  }.recoverWith {
    case tryResult => driver.close().recoverWith {
      case _ => Future.fromTry(tryResult) 
    }
  }
} yield result
```

docs/src/main/tut/docs/alternative_effects.md
BalmungSan: Since these are mainly style issues, I do not think we could get into a agreement. And since it is documentation, I wouldn't give it too much importance, since everyone will have his/her own flavor.

For example, IMHO, the `()` are important, since all this actions are side-effects _(even if they are now pure values, it is my way to immediately distinguish from pure functions and IO wrapped side effects)_.
Also, I have come to really avoid the `_.method` syntax, no matter how _"obvious"_ it is on what I am calling the method, when i have to come back to the code, it always takes me some time to figure it out _(maybe my only exception are things like  `c.foldLeft(M.empty)(_ |+| _)`)_.

But again, those are just my preferences, not need to agree on these.
However, I do think the examples are a little bit easier to read as they are.
_(but since I wrote them, I really can not vote for that 😅)_

@dimafeng What do you think?

docs/src/main/tut/docs/alternative_effects.md
gvolpe: Not sure what's idiomatic with `Future` tbh...

docs/src/main/tut/docs/alternative_effects.md
gvolpe: I agree it's mainly a style diff. 

I disagree in believing `()` is a *side effect*. It shouldn't be in these cases and that's why I prefer to design the interface without it as in

```scala
trait Driver[F[_]] {
  def close: F[Unit] // No () here
}
```

`()` represents a *side effect* when the API is impure IMO.

core/src/main/scala/neotypes/internal/utils/travserse.scala
```diff
@@ -0,0 +1,56 @@
+package neotypes
+package internal.utils
+
+private[neotypes] object traverse {
```
gvolpe: There's a type in the name of the file `travserse` :)

BalmungSan: Oh god, not again 😞 
Thanks for the catch 😄

docs/src/main/tut/docs/alternative_effects.md
BalmungSan: Uhm... I have been checking other libraries like `fs2` and you are right, it should be without the parenthesis. I have to admit it looks awkward to me, but prefer to follow conventions.
Will make this changes as well as the others after @dimafeng leaves his feedback to make them all at once.

gvolpe:
I've been thinking about a step forward of #38 . Since `Driver` and `Session` are resources it would be great if the library handles them as such. But that would mean either introduce a new `Resource / Managed / Bracket` typeclass to abstract over the different libraries or define an interface and have an implementation per module (one for `cats`, another for `zio`, etc).

Here's a sketch of how that might look like:

```scala
final class Driver[F[_]](driver: NDriver)(implicit F: Async[F]) {
  def session: ResourceT[F, Session[F]] =
    session(accessMode = AccessMode.READ)

   def session(accessMode: AccessMode, bookmarks: String*): ResourceT[F, Session[F]] =
    ResourceT.make(
      new Session(
        bookmarks match {
          case Seq()         => driver.session(accessMode)
          case Seq(bookmark) => driver.session(accessMode, bookmark)
          case _             => driver.session(accessMode, bookmarks.asJava)
        }
      )
    )
}

object GraphDatabase {
  def driver[F[_]: Async](uri: String): ResourceT[F, Driver[F]] = ???
}
```

This means we introduced a monad transformer `ResourceT` and in order to make it work we need something similar to the [Bracket](https://github.com/typelevel/cats-effect/blob/master/core/shared/src/main/scala/cats/effect/Bracket.scala) typeclass as a constraint on `F`.

When using `cats-effect` it doesn't feel good using `Resource.make` for `Driver` and `Session`. These should already be treated as resources that's why I'm raising this concern.

If you would like to consider something like this I can open an issue to discuss this further and maybe work it out in a future PR.

dimafeng:
@BalmungSan @gvolpe  sorry guys, I'll do my best to catch up by tommorow

BalmungSan:
@gvolpe Well, IMHO, it definitively should be outside of this **PR** / **Issue**.
Thus, I think the best would be to move that comment to a new **Issue**.

Now, without going in many details, I am not sure if we should really do that.
I do understand your feeling _(since I also use this with `IO`)_, but I think it is _"OK"_ for us as the users to do that. Specially since for things like `Future` it does not make sense and we probably would end up implementing a fake Resource for that.
And, given this is usually done only one time and it is not too long to implement, I do no think the cost overweight the benefits.
Also, the more typeclasses & functions we define, the more constraints we add to implement new `Effects`.

However, I think this deserves further discussion.

gvolpe:
That's fair enough @BalmungSan , just wanted to mention it for possible future consideration but definitely doesn't make any sense for `Future`. I guess that's the trade off of supporting pure and impure effect types.

docs/src/main/tut/docs.md
dimafeng: "alternative effects" sounds good to me

docs/src/main/tut/docs/alternative_effects.md
dimafeng: I don't have strong opinion on `()`. I'd follow practices from **fs2** & **http4s**

dimafeng:
@BalmungSan I almost finished reviewing. Overall looks great but I have two minor concerns:
1. `GraphDatabase` duplicates neo4j `Driver`'s methods which will require maintenance on our side (if neo4j changes api) 
2. 
```
val program = for {
  driver <- GraphDatabase.driver[Future]("bolt://localhost:7687", AuthTokens.basic("neo4j", "****"))
  result <- driver.readSession { session =>
    "match (p:Person {name: 'Charlize Theron'}) return p.name".query[String].single(session)
  }.recoverWith {
    case tryResult => driver.close().recoverWith {
      case _ => Future.fromTry(tryResult) 
    }
  }
} yield result
```
I'm afraid this example will scare scala `Future` users. I suspect that extra complexity in Future use-cases will make the library less attractive to Play Framework, Akka, ... users. 

What do you think?

BalmungSan:
> "GraphDatabase duplicates neo4j Driver's methods which will require maintenance on our side (if neo4j changes api)".

Yeah, that is a problem. But I do think it is a problem we have to carry on _(the same thing happens with the `session` methods on **Driver[F[_]]**)_. Ideally, people should just find here the same functionalities as in the Java driver. _(at least in terms of **Driver** / **Session** creation)_.

But I am open to alternatives, I tried to make the code as SDRY as possible _(SDRY: "Sometimes DRY", meaning just repeat enough to have both code clarity and ease of change)_.

> "I'm afraid this example will scare scala Future users...".

_"Well yes, but actually no"_ 🤣 

I agree with @gvolpe that the official examples should be safe, period!
I would like to know if there is a prettier / idiomatic / simpler way of rewriting that example, as I already said, apart from the tests in this library, I have never used `Future` before.
Anyways, I suspect people using those libraries should already know about that and know how to handle such cases.

We may leave an issue open with the `help wanted` tag, for better examples with `Future` & `Akka`.

BalmungSan:
> ⚠️ Big changes! _(again)_ ⚠️ 

I just pushed a new commit which integrates the concept of **Resource** into **Async**. 🎉
It seems to work well for both `Future` & for `IO`.
_(it would be good to test the snapshot release after merge in some projects, I will do with mine 🔥)_

At the end, I used `Future[A]` as the _resource_ of `Async[Future]` for two reasons:
1. _(80%)_ because **Transaction** should also be a resource, but it has to be async, thus `Id` didn't worked.
2. _(20%)_ even if it is not a resource in the formal sense, at lest it catches exceptions and lets its users to `recover` from it _(which `Id` would not, obviously)_.

Also, I leave the `Future` examples as they were, but I added a note explaining that the examples are unsafe _(unless used with `IO` & friends)_. I still think it would be good to open a new issue with the _"help wanted"_ tag, for better `Future` & `Akka` examples.

Now, I am falling asleep and do not know what else to say lol. Hope you guys like it. 😄

@dimafeng @gvolpe @sloshy

dimafeng:
@BalmungSan this PR is getting bigger 😁 I'll try to complete reviewing tomorrow

BalmungSan:
Just for the record, I tested this PR against the two projects in which I am using **neotypes** right now. And everything seems to work!

I used the following modules:

+ core
+ cats-effect
+ fs2
+ refined

core/src/main/scala/neotypes/Session.scala
```diff
@@ -1,40 +1,45 @@
...
+  def transaction[R[_]](implicit F: Async.Aux[F, R]): R[Transaction[F]] =
+    F.resource(createTransaction) { transaction => transaction.close }
 
-  def close(): F[Unit] =
+  private[neotypes] def createTransaction(implicit F: Async[F]): F[Transaction[F]] =
     F.async { cb =>
       session
-        .closeAsync()
-        .accept { _: Void => cb(Right(())) }
+        .beginTransactionAsync()
```
gvolpe: I thought methods with empty parens were gone?

core/src/main/scala/neotypes/Session.scala
```diff
@@ -1,40 +1,45 @@
...
       } yield res
 
-      result.recoverWith { case ex =>
-        for {
-          _ <- t.rollback()
-          res <- F.failed[T](ex)
-        } yield res
+      result.recoverWith {
+        case ex =>
+          for {
```
gvolpe: I think this is way more idiomatic without for-comprehension (and it's also a one-liner):

```scala
t.rollback *> F.failed[T](ex)
```

Oh... And it's more performant if you're not using `better-monadic-for`.

core/src/main/scala/neotypes/Session.scala
```diff
@@ -1,40 +1,45 @@
...
+            _ <- t.rollback
+            res <- F.failed[T](ex)
+          } yield res
       }
     }
+
+  def close(implicit F: Async[F]): F[Unit] =
+    F.async { cb =>
+      session
+        .closeAsync()
```
gvolpe: Empty parens? (there are any others spread around as well)

dimafeng:
@gvolpe 
> Is this getting merged anytime soon? :)
I hope I'll do it today.

core/src/main/scala/neotypes/Session.scala
BalmungSan: Those are from the **Java** API, as such, those are real side-effects!

core/src/main/scala/neotypes/Session.scala
BalmungSan: Again, that is a real side-effect of the **Java** API.

core/src/main/scala/neotypes/Session.scala
BalmungSan: You are right. Sadly we do not have `*>`. But I may just write the `flatMap`.

core/src/main/scala/neotypes/Session.scala
gvolpe: Fair enough! :smile:

BalmungSan:
@dimafeng Uhm... Apparently **AkkaStreams** is broken now, it is failing randomly, and I really do not understand why...

It seems the transaction / session is being closed before the tests executes.
Maybe it is related to uncontrolled order of execution of `Futures`?
Maybe, is that now every operation requires a new instance of `Async[Future]` instead of reusing the same one?

dimafeng:
@BalmungSan seems that issue was introduced initially. I wasn't sure in my initially akka stream implementation and hit similar issue when was trying to use it from akka-http. I'll take a look at akka-stream module on the weekend. Meanwhile, I can merge this PR since it's not the cause.  Does it sound good to you?

BalmungSan:
For me yes, also it would be great if you could do a release on the weekend, since I am needing it for my projects.

Since I do not use neither `Future` nor `Akka` I really do not care too much for them, so I think it could be fixed latter.

gvolpe:
+1 to a new release (it could be a milestone or RC).

dimafeng:
🎉

PR #40, opened 2019-07-27
Add mapping Relationship to case classes in ResultMappers

pbylicki:
Hi,
I've had a case where I wanted to map relationship from a query result to a case class and found out that it was not yet possible so I decided to add it :)


pbylicki:
@dimafeng There is one Akka Streams related test that is failing. It seems to be unstable when running locally and I see that it failed in some of the recent Travis builds too. Can you try to rerun the build or advise how to proceed?

dimafeng:
@pbylicki awesome! looks great!
@BalmungSan thanks for being on top of this! Akka stream is still on my todo list (unfortunately, have no free time these days)

PR #43, opened 2019-08-10
WIP: Fix transaction

### LINKED ISSUE #42: Discussion: Transaction as a Resource
Ok so, back on #39 I made `Transaction[F]` a **Resource**. Thus, the `transaction` method on `Session[F]` returns an `R[Transaction[F]]`. That had the negative consequence that we couldn't use `Id` as the **Resource** of `Future`s. But, it seemed like the best thing to do.
However, I have been thinking about it, and I now believe I was wrong!

First, it is true that a **Transaction** should always be `close`d at the end _(that was the main reason for making it a **Resource**)_.
But, the problem is that it has two ways of being closed: `commit`ted & `rollback`ed.

Second, our `close` implementation tries to `commit` the transaction and if that fails, then it tries to `rollback`. At first, that made sense for me...

> I thought about a case like, creating multiple nodes in the same transaction, but if one of the nodes breaks an unique constraint, the commit would fail, and the rollback seemed reasonable.

... However, the Java driver already do this. If a transaction fails to be committed, nothing will happen. And, calling `commit` will return an exception explaining the problem.

Third, If one wants to manually rollback the transaction _(for any reason)_. Then when the resource `close`s and calls `commit`, it will throw an exception! _(because the transaction was already rollbacked and closed)_.

Note: All these apply only to `IO` _& friends_ TM - `Future` does not have this problem, as its **Resource** is not doing nothing.

------

Thus, my question is:

Should we make a **Transaction** just shared state. instead of a **Resource**. That means, returning `F[Transaction[F]]` instead?
Or maybe, should we keep some internal _(mutable)_ state, that changes if the transaction was `rollback`ed, and check that on the close operation?

Do anyone has any other idea or opinion about this?
I am specially interested in hearing about @gvolpe & @dimafeng

BalmungSan:
Closes #42 

Ok so, as we said, this basically reverts the change of treating **Transactions** as **Resources**.
This has two consequences:

1. We can use `Id` as the **Resource** of `Futures` again! - Thus, no more `Future[Driver[Future]]`, just `Driver[Future]`.
2. The user has the ability _(and the obligation)_ of calling either `commit` or `rollback` in order to finish a transaction and being able to start a new one.

----------

Right now I am missing the update of the documentation....

... and, I am still thinking in what would be the best way of having a way of returning a `Resource[F, Transaction[F]]`, only for _IO & Friends TM_, that `commit`s as it's close action.


dimafeng:
@BalmungSan looks good to me. There's some problems with the build (not akka-stream related), is it something expected?

@gvolpe could you please take a look if you have some spare time?

BalmungSan:
@dimafeng yeah, I believe I fixed the AkkaStreams issue. But I do not understand the other failure, however I made a final change an forgot to test it before pushing. I hope it is easy to fix but I have not had the time to finish it. I believe it would be tomorrow.
Anyways, please do not merge it while it has the `WIP` in the name :D _(applies to future PRs too)_.

@gvolpe I would really appreciate if you can have a look!

gvolpe:
I'm not very familiar with the internals but what I see looks good to me :+1: 

I think the docs need some update right? :smile:

BalmungSan:
@gvolpe Yes. That is mainly the reason this is still a WIP _(plus the unexpected failure)_.

BalmungSan:
Okey, fixed the bug, it took me a while _(specially because I couldn't make it work as how I would like)_.

I am still missing the documentation and the _auto-commitable_ **Transaction**.
I would expect to be working on that during the week.

BalmungSan:
Great news, I came up with an idea to have a way to auto commit transactions.

I added a new method to our `Async` **typeclass**:

```scala
def guarantee[A, B](fa: F[A])(f: A => F[B])(finalizer: A => F[Unit]): F[B]
```

_(It is somewhat similar to a `bracket`)_

For **Future** we have our own implementation, that is exactly the same we have been using for things like `readSession`.
And for **IO & friends** it fallback to using `Resource.use`, which give us very strong guarantees!

Thus, right now we will have:

+ `query.execute(session)` - If you want to execute only one query and have it auto commited.
+ `session.transact { tx =>... }` - If you want to execute one or more queries in the same transaction, and have all them auto commited.
+ `val tx = session.transaction` - If you want to execute one or more queries in the same transaction, but you need to control when to commit / close.

With that I am done with code changes, but I am still missing documentation.
I would expect to do that this sunday or next monday.

-----

PS: Apparently AkkaStreams is still broken somehow... I thought the problem was my change to **Transaction** but apparently the problem was not that. 😞 
Also, sorry for the false positive of closing this, I deleted the branch by accident. 😅

core/src/main/scala/neotypes/Session.scala
```diff
@@ -17,21 +17,7 @@ final class Session[F[_]](private val session: NSession) extends AnyVal {
...
-      result.recoverWith {
-        case ex: Throwable =>
-          tx.rollback
-            .flatMap(_ => F.failed[T](ex))
-            .recoverWith {
-              case _ => F.failed[T](ex)
-            }
-      }
-    }
+    transaction.guarantee(txF)(_.commit)
```
dimafeng: Not sure if I'm fully following the flow but it seems we always commit regardless of `txF` result. 
```scala
s.transact {tx =>
  for {
    res <- c"create (p:Record {personId: 1234)".query[Unit].execute(tx)
    _ <- Future.failure(new Exception("ERROR"))
  } yield ()
}
```

dimafeng: Seems we'll need to add transaction tests like that one at some point

dimafeng:
@BalmungSan yeah, AkkaStreams are on me! Hopefully I'll have some time in a week

core/src/main/scala/neotypes/Session.scala
BalmungSan: @dimafeng Agree, I will add some tests to the transactions, just to make sure they work as intended for each effect.

Answering your first question, yes it will commit, regardless what happens.
My logic is that, on the scope of a `transact` there should only be logic related to **Neo4j** operations / queries. As such, the only possible errors, are those related with transactions themselves. And in those cases, the transaction will be automatically rollbacked before the commit, and at the end you will have the original exception _(the tests should confirm that!)_.

If that is not the case, and you have custom logic that may also fail. Then, `transact` is not the abstraction you should be using but `session.transaction.flatMap { tx => ??? }` and then a manual rollback or commit.

I am open to discussion if you believe the previous case is common enough... But, then the implementation becomes very tricky.

dimafeng: I actually don't have strong opinion on this one. I'm open to the most expected (industry standard) behavior - I guess it's a question of what expected / industry standard behavior is 😁 

I think we can ask other people. cc @sloshy @gvolpe 

Maybe a stupid question, but I don't have answer off the top of my head - what would happen in case of:
```
s.transact {tx =>
  for {
    _ <- c"create (p:Record {personId: 1234)".query[Unit].execute(tx)
    _ <- c"broken cypher query".query[Unit].execute(tx)
  } yield ()
}
```

BalmungSan: Everything will be rollbacked by the **Neo4j java driver**. Thus, the moment we try to call `commit`, it will fail.

Again, the tests should confirm that, and that we preserve the original exception.
If that is not the case, then yes, we would need to change it.

BalmungSan:
@dimafeng 

Ok, I managed to have `rollback` for any kind of **failure** and `commit` if everything is fine.
Also, I just added some tests that confirm the behavior is as expected.

I am still missing the documentation, probably this weekend I will have the time.

dimafeng:
@BalmungSan looks great!

BalmungSan:
@dimafeng I am done! 🎉 

Please, give it a good review _(especially to the documentation, written English is not my best skill)_.

If you see everything ok, please merge it and publish a new version as soon as you can.
:+1:

docs/src/main/tut/docs/parameterized_queries.md
```diff
@@ -10,23 +10,22 @@ title: "Parameterized Queries"
...
 
-assert(query.rawQuery == "create (a:Test {name: $p1})")
+assert(query.rawQuery == "CREATE (a:Test {name: $p1})")
 assert(query.params == Map("p1" -> QueryParam("John")))
 ```
 
 All parameters will be converted to neo4j supported types (please see [Supported types](types.html)).
 
-`c` interpolator creates `DeferredQueryBuilder` which is an immutable representation of a cypher query.
+`c` interpolator CREATEs `DeferredQueryBuilder` which is an immutable representation of a cypher query.
```
dimafeng: `CREATEs` seems an auto replacement 😁

dimafeng:
@BalmungSan just one minor comment. The rest looks great!

docs/src/main/tut/docs/parameterized_queries.md
BalmungSan: Oh...
Yeah, you catch me there, sorry.
There were some others, it seems I forget this one.
Let me fix it real quick!

BalmungSan:
@dimafeng I just patched the last commit with the fix.
Everything should be ready to go!

dimafeng:
@BalmungSan I've just released it

PR #51, opened 2019-10-05
Spelling corrections

bengraygh:
This PR corrects a number of spelling issues in the project and makes no functional changes.

For example:
UncoercibleException -> IncoercibleException
CompabilityMappers -> CompatibilityMappers
commited -> committed

All tests pass.

### DIFF
--- a/core/src/main/scala/neotypes/mappers.scala
+++ b/core/src/main/scala/neotypes/mappers.scala
@@ -1,6 +1,6 @@
 package neotypes
 
-import exceptions.{PropertyNotFoundException, UncoercibleException}
+import exceptions.{PropertyNotFoundException, IncoercibleException}
 import types.QueryParam
 
 import org.neo4j.driver.v1.Value
@@ -285,7 +285,7 @@ object mappers {
             try {
               Right(f(v))
             } catch {
-              case ex: Uncoercible => Left(UncoercibleException(ex.getLocalizedMessage, ex))
+              case ex: Uncoercible => Left(IncoercibleException(ex.getLocalizedMessage, ex))
               case ex: Throwable   => Left(ex)
             }
 
--- a/cats-data/src/main/scala/neotypes/cats/data/CatsData.scala
+++ b/cats-data/src/main/scala/neotypes/cats/data/CatsData.scala
@@ -2,7 +2,7 @@ package neotypes
 package cats.data
 
 import internal.utils.traverse.{traverseAs, traverseAsList, traverseAsMap, traverseAsSet, traverseAsVector}
-import exceptions.{PropertyNotFoundException, UncoercibleException}
+import exceptions.{PropertyNotFoundException, IncoercibleException}
 import mappers.{ParameterMapper, ResultMapper, ValueMapper}
 import types.QueryParam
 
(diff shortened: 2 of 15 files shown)


core/src/main/scala/neotypes/mappers.scala
```diff
@@ -285,7 +285,7 @@ object mappers {
             try {
               Right(f(v))
             } catch {
-              case ex: Uncoercible => Left(UncoercibleException(ex.getLocalizedMessage, ex))
+              case ex: Uncoercible => Left(IncoercibleException(ex.getLocalizedMessage, ex))
```
dimafeng: `Uncoercible` comes from neo4j driver and the original idea was to have them named consistently. 

@BalmungSan what do you think about it?

BalmungSan: I did a little research _(since English is not my native language)_ and it seems that _"Uncoercible"_ does not exists and _"Incoercible"_ is the appropriate expression...

However, as @dimafeng said, the original **neo4j** exception is _"Uncoercible"_.
Nevertheless, that name only exists in the **Java** driver, the **C#** and the **Python** use different names.

Thus, IMHO, it would be better to have proper name. I would suggest remove that and just use `ConversionException` just to avoid confusion.
What do you all think?

bengraygh: It is a shame some consistency is lost for the sake of correctness. ConversionException sounds very clear to me as an alternative. Happy to update the PR with whatever the consensus is here.

dimafeng: > Thus, IMHO, it would be better to have proper name. I would suggest remove that and just use ConversionException just to avoid confusion.
What do you all think?

Good idea. 

Taking into account:

> Nevertheless, that name only exists in the Java driver, the C# and the Python use different names.

add

> "Uncoercible" does not exists 

plus the fact that `UncoercibleException` takes a cause of the exception, I'd make the following change:

1. Rename `UncoercibleException` -> `IncoercibleException` as it's done in this PR (no changes needed)
2. Make `IncoercibleException` extend `ConversionException` so that the use of `ConversionException` is sufficient in most cases.

@BalmungSan would it make more sense?
@bengraygh if @BalmungSan agrees with the proposal, would you be willing to make p. 2 in the scope of this PR (if not that's totally fine, we can make it as a separate PR)?

BalmungSan: I like the idea
@bengraygh we'll wait for your answer.

bengraygh: This sounds good. I'm not 100% clear on what (2) entails because ConversionException is a concrete implementation of NeotypesException which doesn't take a 'cause'. What did you have in mind?

Either way, happy to do the changes to this PR.

dimafeng: I started thinking about it, and it doesn't seem to be good idea anymore 😁 

Changing `ConversionException` as follows
```scala
case class ConversionException(message: String, cause: Option[Throwable] = None) extends NeotypesException(message, cause)
final case class IncoercibleException(message: String, cause: Uncoercible) extends ConversionException(message, cause)
```
would certainly make api design ugly and potentially will break destructuring/pattern matching.

@BalmungSan if you don't have any other comments, I'd like merge this PR as-is.

BalmungSan: Uhm... yeah I forgot those were case classes.
Lest just leave it as how it is right now _(in this PR)_.

dimafeng:
@bengraygh thank you!

PR #55, opened 2019-12-05
Add scaladoc

geoffjohn11:
What do you think of this content for AkkaStreams?  I started here and thought I would get some feedback before continuing.

### DIFF
--- a/akka-stream/src/main/scala/neotypes/akkastreams/AkkaStreams.scala
+++ b/akka-stream/src/main/scala/neotypes/akkastreams/AkkaStreams.scala
@@ -4,11 +4,31 @@ import akka.stream.scaladsl.{Flow, Source}
 
 import scala.concurrent.{ExecutionContext, Future}
 
+/**
+  * neotypes Akka Streams
+  * @see <a href = https://neotypes.github.io/neotypes/docs/streams.html#akka-streams-neotypes-akka-stream>AkkaStreams example</a>
+  */
 trait AkkaStreams {
+
+  /** neotypes support for Akka Streams
+    *
+    * @param ec implicit scala.concurrent ExecutionContext that will be passed down to all Akka Streams operations
+    * @return neotype AkkaStream with effect type as the scala.concurrent Future
+    */
   implicit final def akkaStream(implicit ec: ExecutionContext): neotypes.Stream.Aux[AkkaStream, Future] =
     new neotypes.Stream[AkkaStream] {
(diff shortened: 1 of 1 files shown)


akka-stream/src/main/scala/neotypes/akkastreams/AkkaStreams.scala
```diff
@@ -4,11 +4,30 @@ import akka.stream.scaladsl.{Flow, Source}
...
 import scala.concurrent.{ExecutionContext, Future}
 
+/**
+  * neotype Akka Streams
+  */
 trait AkkaStreams {
+
+  /** neotype support for Akka Streams
+    *
+    * @param ec implicit scala.concurrent ExecutionContext
```
BalmungSan: It would be good to explain why we need an implicit execution context.
Something along the lines of:

_"The execution context that will be passed down to all **Akka Streams** operations"_.

dimafeng:
>@dimafeng May you help us in that? Maybe we could start to publish the scaladoc as it is right now?
@BalmungSan good question. I suspect it can be easily published along with the site as it's just static html. I'll take a look over the weekend

akka-stream/src/main/scala/neotypes/akkastreams/AkkaStreams.scala
```diff
@@ -4,11 +4,30 @@ import akka.stream.scaladsl.{Flow, Source}
 
 import scala.concurrent.{ExecutionContext, Future}
 
+/**
+  * neotype Akka Streams
```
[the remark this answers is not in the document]
geoffjohn11: I added a link to the documentation and changed the typo.

akka-stream/src/main/scala/neotypes/akkastreams/AkkaStreams.scala
geoffjohn11: I changed the description.

dimafeng:
@BalmungSan this PR looks good to me

BalmungSan:
@dimafeng yeah to me too. But I do not know if @geoffjohn11 prefers to gets this one merged now or rather add more before.

geoffjohn11:
I'm adding doc to DeferredQuery now.  Would you like to get that into this pr?

BalmungSan:
@geoffjohn11 IMHO the best would be to make one PR per section, so I will merge this one.
Thanks!

PR #56, opened 2019-12-09
DeferredQuery scaladoc

geoffjohn11:


geoffjohn11:
@dimafeng @BalmungSan initial feedback back on doc welcome.

core/src/main/scala/neotypes/DeferredQuery.scala
```diff
@@ -8,63 +8,187 @@ import scala.collection.compat.Factory
...
 
+  /** Evaluate query for effect type F
+    *
+    * @param tx neotypes transaction for effect type F
+    * @param F effect type F
+    * @param rm result mapper for type T
+    * @tparam F effect type
+    * @return F[T]
+    */
   def execute[F[_]](tx: Transaction[F])
```
geoffjohn11: @BalmungSan Does the actual execution of a statement depend on when (for example in Cats Effect) an unsafe operation triggers the side effect to occur?  In that case is it impossible to know when the actual effect will be executed?  I'm having difficulty describing what an operation like execute performs when (I think) the output of this method is a description of the side effect to be evaluated, not the execution itself.

core/src/main/scala/neotypes/DeferredQuery.scala
```diff
@@ -95,9 +219,24 @@ private[neotypes] object DeferredQuery {
...
+  * val stringInter = c"""create (a:Test {name: $name,""" + c"born: $born})"
+  * val deferredQuery = stringInter.query
+  * </pre>
+  *
+  * @param parts
+  */
 final class DeferredQueryBuilder private[neotypes] (private val parts: List[DeferredQueryBuilder.Part]) {
   import DeferredQueryBuilder.{PARAMETER_NAME_PREFIX, Param, Part, Query}
 
+  /** Creates a DeferredQuery from the query and params
```
BalmungSan: Creates a deferred query from this builder.

```scala
@tparam T The result type of the constructed query.
```

core/src/main/scala/neotypes/DeferredQuery.scala
```diff
@@ -142,9 +281,19 @@ final class DeferredQueryBuilder private[neotypes] (private val parts: List[Defe
       )
   }
 
+  /** Add together the Query and Param of two DeferredQueryBuilders
```
BalmungSan: For this one and the other that receives a string, it would be better to say that it _concats_ together two queries. Instead of saying it adds.

core/src/main/scala/neotypes/DeferredQuery.scala
BalmungSan: I wouldn't bother too much with when the real execution is happening, but rather on what returned **effect** describes. In this case, as I said in the other comment, it is:

> Executes the query and ignores its output.

geoffjohn11: @BalmungSan Sounds good, made the changes with examples

dimafeng:
Look good to me! Great job!

PR #57, opened 2019-12-12
driver documentation

geoffjohn11:


core/src/main/scala/neotypes/Driver.scala
```diff
@@ -8,14 +8,42 @@ import org.neo4j.driver.v1.{AccessMode, Driver => NDriver}
...
+  * @tparam F effect type for driver
+  */
 final class Driver[F[_]](private val driver: NDriver) extends AnyVal {
+
+  /** acquire session to a database in read mode
+    * @note a session is a resource that needs to be allocated and released
+    *
+    * @param F asynchronous effect type with resource type defined
+    * @tparam R effect type that allows creating, using, and releasing a resource
+    * @return Session[F] in effect type R
```
BalmungSan: `R` is not really an **effect type** it is rather a **resource type** which is dependent of the `F` **effect type** which was used for the creation of the Driver.

Also, we should be very careful when talking about this as since for **Future** the _"resource"_ type is the **Id** meaning there isn't really a resource.

core/src/main/scala/neotypes/Driver.scala
```diff
@@ -8,14 +8,42 @@ import org.neo4j.driver.v1.{AccessMode, Driver => NDriver}
...
+    * @param bookmarks bookmarks passed between transactions for neo4j casual chaining
+    * @param F asynchronous effect type with resource type defined
+    * @tparam R effect type that allows creating, using, and releasing a resource
+    * @return Session[F] in effect type R
+    */
   def session[R[_]](accessMode: AccessMode, bookmarks: String*)
                    (implicit F: Async.Aux[F, R]): R[Session[F]] =
     F.resource(createSession(accessMode, bookmarks))(session => session.close)
 
+  /** create a session from the neo4j driver
```
BalmungSan: This is a private method, it is not needed to be documented as it won't be visible.

core/src/main/scala/neotypes/Driver.scala
```diff
@@ -8,14 +8,42 @@ import org.neo4j.driver.v1.{AccessMode, Driver => NDriver}
...
 import scala.language.higherKinds
 
+/** neotypes driver for accessing a neo4j graph database
+  *
+  * @param driver neo4j driver
+  * @tparam F effect type for driver
+  */
 final class Driver[F[_]](private val driver: NDriver) extends AnyVal {
+
+  /** acquire session to a database in read mode
```
dimafeng: A minor comment about capitalization of first letter. It would be great to make it consistent with previous PRs

core/src/main/scala/neotypes/Driver.scala
dimafeng: Yeah, I understand your concern here. I'd say *resource type* works for me and maybe we can add a note that would emphasize on this exception with *Future*

BalmungSan: Yeah seems like the best way.

geoffjohn11:
Thanks for the comments @BalmungSan and @dimafeng.  The doc writing is a really good way to learn about the functional programming and neo4j.  Hoping once it's done I'll be in a position to contribute something a bit more meaningful too.

BalmungSan:
@geoffjohn11 Never underestimate the meaningfulness of documentation!

But, I get your point, looking forward to that. If you have any doubts or want to discuss something do not hesitate in talking to us on the gitter channel.

core/src/main/scala/neotypes/Driver.scala
```diff
@@ -8,10 +8,36 @@ import org.neo4j.driver.v1.{AccessMode, Driver => NDriver}
 import scala.jdk.CollectionConverters._
 import scala.language.higherKinds
 
+/** A neotypes driver for accessing the neo4j graph database
+  * A driver wrapped in the resource type can be created using the neotypes GraphDatabase
+  * {{{
+  *    val driver: F.R[Driver[F]] = GraphDatabase("bolt://localhost:7687")(F)
```
BalmungSan: The intended usage is:

```scala
val driver = GraphDatabase.driver[F]("bolt://localhost:7687")
```

core/src/main/scala/neotypes/Driver.scala
```diff
@@ -8,10 +8,37 @@ import org.neo4j.driver.v1.{AccessMode, Driver => NDriver}
...
 
+/** A neotypes driver for accessing the neo4j graph database
+  * A driver wrapped in the resource type can be created using the neotypes GraphDatabase
+  * {{{
+  *    val driver = GraphDatabase[F]("bolt://localhost:7687").driver
+  * }}}
+  *
+  * @param driver neo4j driver
+  * @tparam F effect type for driver
+  * @note the resource type for scala.Future has no failure support, see <a href = https://neotypes.github.io/neotypes/docs/alternative_effects.html>alternative effects</a>
```
BalmungSan: I would rather write something like this:
_"When your effect type is `scala.Future` there is no concept of **Resource**. For more information see ...."_

core/src/main/scala/neotypes/Driver.scala
BalmungSan: I bit tedious... But, it would be worth adding the same note about resources and future here.

But then, it would be good adding that note on all the methods that return a resource, and there are many.
It would be good if we could have the message in one place and reference it in many, as a variable, so we do not have to repeat ourselves and making changes in a future would be easier.

I am pretty sure that exists. I believe to remember seeing something like that on the official scaladoc, precisely on the collections framework.
Could you help us looking for that? It would be really helpful!!!

BalmungSan: @geoffjohn11 Pretty cool!! Thank.

But now the thing is, that note will be also used in the **Session** class and in the **GraphDatabase** object.
You know if there is a way to have that `@define` in a global space? _(or it is already global)_. Maybe we could put it in another file, like `globals_doc.scala`?

PR #60, opened 2020-01-10
case class mapper for maps

### LINKED ISSUE #26: Feature: case class mapper for maps
If using [`map projections`](https://neo4j.com/docs/cypher-manual/current/syntax/maps/#cypher-map-projection), **neotypes** is unable to map the returned values to case classes. A common use case I have _(which I believe could be useful for others)_ is having a node with a lot of fields to which you want to add some other computed fields and return that as a case class.
Example:

```scala
c"""MATCH (u: User { name: $name })
      ...
     RETURN u { .*, score: sum(???) }
    ORDER BY u.score
    LIMIT 10""".query[ScoredUser].list(session)
```

Right now we are solving the problem like this:

```scala
c"""MATCH (u: User { name: $name })
      ...
     RETURN
       u.name AS name,
       ...
       sum(???) AS score
    ORDER BY u.score
    LIMIT 10""".query[ScoredUser].list(session)
```

Not a big deal, but is kind of tedious to add every property to the return.

------

**Disclaimer:** I do not know anything about shapeless, so I can not help with this. Being honest, I do not even know if this is easy, doable, difficult or impossible lol 😅 
But I guess I do not lose anything opening this just to see if anyone knows how to do it.

geoffjohn11:
closes #26

### DIFF
--- a/core/src/main/scala/neotypes/implicits/mappers/ResultMappers.scala
+++ b/core/src/main/scala/neotypes/implicits/mappers/ResultMappers.scala
@@ -1,18 +1,17 @@
 package neotypes
 package implicits.mappers
 
-import java.time.{Duration, LocalDate, LocalDateTime, LocalTime, Period, OffsetDateTime, OffsetTime, ZonedDateTime}
+import java.time._
 import java.util.UUID
 
-import mappers.{ResultMapper, TypeHint, ValueMapper}
-import types.Path
-
+import neotypes.mappers.{ResultMapper, TypeHint, ValueMapper}
+import neotypes.types.Path
 import org.neo4j.driver.internal.types.InternalTypeSystem
 import org.neo4j.driver.internal.value.IntegerValue
 import org.neo4j.driver.v1.Value
-import org.neo4j.driver.v1.types.{Entity, IsoDuration, Node, Path => NPath, Point, Relationship}
-import shapeless.{:: => :!:, HList, HNil, LabelledGeneric, Lazy, Witness, labelled}
+import org.neo4j.driver.v1.types.{Entity, IsoDuration, Node, Point, Relationship, Path => NPath}
 import shapeless.labelled.FieldType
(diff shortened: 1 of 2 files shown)


geoffjohn11:
The other use case of Tuple to Map isn't yet complete, but this one had a bit less to it.  Those test cases are still commented out in MapSessionSpec.  Any thoughts on this?

geoffjohn11:
> Hi @geoffjohn11 Thanks for this!!
> 
> Could you make sure that you enable the _"allow changes from maintainers"_ option in this PR, I would like to make a simple style change and ensure that `collectMapFields` is _tail-recursive_ before merging.

That option is checked, so you should be able to make edits.

collectMapFields is not tail-recursive, so the annotation will cause a compilation failure, but I can take a look at making it tailrec.

BalmungSan:
@geoffjohn11 I just remembered that **Neo4j** doesn't support nested maps, thus your recursive step wasn't necessary.

When I removed that, I discovered that an [**Entity**](https://neo4j.com/docs/api/java-driver/current/org/neo4j/driver/v1/types/Entity.html) is basically a **Map**, so we could reuse our old logic _(with some minor tweaks)_.

So, in the end, we didn't need to do anything 😆 
However, if it hadn't been for this PR we probably would never notice that.
Thank you very much for your interest in the project 👍

geoffjohn11:
> @geoffjohn11 I just remembered that **Neo4j** doesn't support nested maps, thus your recursive step wasn't necessary.
> 
> When I removed that, I discovered that an [**Entity**](https://neo4j.com/docs/api/java-driver/current/org/neo4j/driver/v1/types/Entity.html) is basically a [**Map**](https://neo4j.com/docs/api/java-driver/current/org/neo4j/driver/v1/types/MapAccessor.html), so we could reuse our old logic _(with some minor tweaks)_.
> 
> So, in the end, we didn't need to do anything 
> However, if it hadn't been for this PR we probably would never notice that.
> Thank you very much for your interest in the project

Interesting!  Thanks for your help.  I'll try to get the two other test in MapSessionSpec working.

PR #64, opened 2020-01-16
Fix PathSessionSpec

geoffjohn11:
The test was failing because the result set returned was greater than 1
The doc on neo4j singleAsync says:
* Asynchronously return the first record in the result, failing if there is not exactly one record left in the stream.

It was a little tricky to realize this was the problem because the exception being thrown is hidden here:
```case _: NoSuchRecordException => resultMapper.to(Seq.empty, None)```

### DIFF
--- a/core/src/test/scala/neotypes/PathSessionSpec.scala
+++ b/core/src/test/scala/neotypes/PathSessionSpec.scala
@@ -3,6 +3,7 @@ package neotypes
 import neotypes.implicits.mappers.results._
 import neotypes.implicits.syntax.string._
 import org.neo4j.driver.v1.types.{Node, Relationship}
+import collection.JavaConverters._
 import shapeless._
 
 import scala.concurrent.Future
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: core/src/test/scala/neotypes/PathSessionSpec.scala [hunk 1]
```diff
@@ -21,11 +21,9 @@ class PathSessionSpec extends BaseIntegrationSpec[Future] {
...
   }
 
-  it should "assign path to case class field" ignore execute { s =>
-    for {
-      data <- "match path=(_:Person)-[*]->() return { path: path }".query[Data].single(s)
-    } yield {
-      assert(data.path.nodes.size == 2)
+  it should "assign path to case class field" in execute { s =>
+    "match path=(_:Person)-[*]->() return { path: path }".query[Data].list(s).map{
+      res => assert(res.head.path.nodes.size == 2)
```
BalmungSan: We have been having this convention over all the project.

```scala
// If just one named param.
foo.bar { param =>
   body
}

// If using pattern matching.
foo.bar {
  case param =>
     body
}
```

So, in this case, please move the parameter to the end of the previous line and please add a space between `map` and `{`.

geoffjohn11: > We have been having this convention over all the project.
> 
> ```scala
> // If just one named param.
> foo.bar { param =>
>    body
> }
> 
> // If using pattern matching.
> foo.bar {
>   case param =>
>      body
> }
> ```
> 
> So, in this case, please move the parameter to the end of the previous line and please add a space between `map` and `{`.

@BalmungSan made the formatting fixes and also added more assertions to make it clear what the result set returned is.

Issue #44, opened 2019-08-26
Discussion: Session + Concurrency


BalmungSan:
The [**javadoc** for `Session`](https://neo4j.com/docs/api/java-driver/current/org/neo4j/driver/v1/Session.html) explicitly states that:

> "Multiple sessions should be used when working with concurrency; session implementations are not thread safe."
> "At most one transaction may exist in a session at any point in time. To maintain multiple concurrent transactions, use multiple concurrent sessions."

Thus, in a context of a web application, it seems the best would be to have a pool of opened `Sessions` which map one to one with a **thread pool** _(`ExecutionContext`)_ of the same size.

Also, it seems to me that this is something the library should provide out of the box. But it is still not clear to me how this should work. I am only sure the underlying EC should be supplied by the user.

geoffjohn11:
@BalmungSan
I created an example of sharing a neotypes Session across threads (which is easiest to do using the Scala Future effect type).  Since there already was an example project using Akka http and futures in the examples repo, I used that instead of the http4s and IO file you gave me, just so I could get something going as quickly as possible.  
My test repo is located here:
https://github.com/geoffjohn11/sessionShare 

Uncomment the duplicated line in “dangerMethod” to execute the “incrementMovieYear” method simultaneously in different threads (executing in separate Scala Futures).  This results in a neo4j session exception: 
```
org.neo4j.driver.internal.NetworkSession.lambda$ensureNoOpenTx$35(NetworkSession.java:526)

'You cannot begin a transaction on a session with an open transaction; either run from within the transaction or use a different session.'
```
It looks to me that because the neotypes reads and writes all open a transaction with beginTransactionAsync, and 
At most one transaction may exist in a session at any point in time (in the neo4j session doc), the action fails.  The second thread can’t use the shared session.

BalmungSan:
After thinking a while about this and after carefully reading [this](https://neo4j.com/docs/driver-manual/current/cypher-workflow/) I believe the best idea would be what Gabriel suggested in the first place, having some kind of semaphore that ensure each **Session** instance can only be used from a thread a the time. This will guarantee thread safety.
geoffjohn11 proposal of hiding the **Session** also ensured thread safety, however creating and destroying sessions every time a query was made was costly and removed an important property of the system of being casually consistent.

If some downstream user wants to execute queries in parallel, then it would be his/her responsibility to instantiate many **Sessions** and load balance all the petitions between them.

-----

I would try to implement this in this weekend.
[...]

PR #66, opened 2020-01-21
Add SessionConcurrentSpec

geoffjohn11:


geoffjohn11:
@BalmungSan This is an example test of sharing a session in parallel with the cats IO effect type.  Not sure we should merge this, but I wanted to share this example.

BalmungSan:
Cool @geoffjohn11 I will try to pull and play with this code.

BalmungSan:
@dimafeng  This PR Adds a new spec that checks the behavior of using a session concurrently.
For now, it fails. But ideally, we would change this in the future, to ensure the session can be sued concurrently for all effect types _(see #44 )_.

@geoffjohn11  Can you take a look at the changes?

core/src/test/scala/neotypes/FutureConcurrentSessionSpec.scala
```diff
@@ -0,0 +1,14 @@
+package neotypes
+
+import scala.concurrent.Future
+
+class FutureConcurrentSessionSpec extends ConcurrentSessionSpec[Future] {
+  override def fToFuture[T](future: Future[T]): Future[T] =
+    future
+
+  override def runConcurrently(a: Future[Unit], b: Future[Unit]): Future[Unit] =
+    for (_ <- a; _ <- b) yield () // Because Futures are eager, they are already running concurrently.
```
geoffjohn11: This comment is helpful

core/src/test/scala/neotypes/ConcurrentSessionSpec.scala
```diff
@@ -0,0 +1,38 @@
+package neotypes
+
+import neotypes.implicits.mappers.all._
+import neotypes.implicits.syntax.cypher._
+import neotypes.internal.syntax.async._
+import org.neo4j.driver.v1.exceptions.ClientException
+import org.scalatest.compatible.Assertion
+import scala.concurrent.Future
+import scala.reflect.ClassTag
+
```
geoffjohn11: Intellij tells me that you can clean up the imports.

BalmungSan: Will take a look :+1:

PR #67, opened 2020-01-23
WIP encapsulate session

geoffjohn11:


geoffjohn11:
Here is one implementation at encapsulating Session inside Driver.  I'm not sure this is the best way, it's just an idea.

BalmungSan:
@geoffjohn11 Thanks for the proposal.
But, as I mentioned in #44 this solution is not really optimal, since a **Session** ensures eventual changing of the transactions.

Superseded by #163

PR #83, opened 2020-02-20
Add AsyncGuaranteeSpec

geoffjohn11:


geoffjohn11:
@BalmungSan I made a test for guarantee method in futureAsync, since it seemed like might be good to have some testing coverage there.  I also made a small refactor to the method, which is mostly just a stylistic difference.

core/src/test/scala/neotypes/FutureAsyncSpec.scala
```diff
@@ -0,0 +1,92 @@
...
+
+  trait FutureAsyncFixture {
+    var counter = 0
+    var resourceIsOpen = true
+    val fa = Future.successful(1)
+    val f: Int => Future[String] = i => Future.successful(i.toString)
+    val finalizer = { (_: Int, _: Option[Throwable]) => {Future.successful {resourceIsOpen = false}}}
+  }
+
+  "async future" should "execute finalizer" in new FutureAsyncFixture{
```
BalmungSan: Could you please leave a space before the opening curly brace?

core/src/test/scala/neotypes/FutureAsyncSpec.scala
```diff
@@ -0,0 +1,92 @@
...
+    ScalaFutures.whenReady(Async.futureAsync.guarantee(fa)(f)(finalizer)){
+      w =>
```
BalmungSan: Could you give this a variable a better name? I do not understand what `w` is supposed to mean.

core/src/test/scala/neotypes/FutureAsyncSpec.scala
```diff
@@ -0,0 +1,92 @@
...
+    }
+  }
+
+  "async future" should "execute finalizer when fa fails and return the exception" in new FutureAsyncFixture{
+    override val fa = Future.failed(new Exception("Boom!"))
+
+    ScalaFutures.whenReady(Async.futureAsync.guarantee(fa)(f)(finalizer).failed){
+      ex =>
+        ex.getMessage shouldBe "Boom!"
+        resourceIsOpen shouldBe true //resource is closed
```
BalmungSan: I am pretty sure here it should be `false`.

core/src/test/scala/neotypes/FutureAsyncSpec.scala
```diff
@@ -0,0 +1,92 @@
...
+    }
+  }
+
+  "async future" should "execute finalizer when f fails and return the exception" in new FutureAsyncFixture{
```
BalmungSan: What is the difference between this test and the previous one?

core/src/test/scala/neotypes/FutureAsyncSpec.scala
```diff
@@ -0,0 +1,92 @@
...
+    ScalaFutures.whenReady(Async.futureAsync.guarantee(fa)(f)(finalizer).failed){
+      ex =>
+        ex.getMessage shouldBe "Boom!"
+        resourceIsOpen shouldBe true //resource was not closed
+    }
+  }
+
+  "async future" should "return finalizer exception" in new FutureAsyncFixture {
+    override val finalizer = { (_: Int, _: Option[Throwable]) => {
+      counter = counter + 1
```
BalmungSan: `counter += 1`

core/src/test/scala/neotypes/FutureAsyncSpec.scala
```diff
@@ -0,0 +1,92 @@
...
+      Future.failed(new Exception(s"Finalizer failure count: ${counter}!"))
+      }
+    }
+
+    ScalaFutures.whenReady(Async.futureAsync.guarantee(fa)(f)(finalizer).failed){
+      ex =>
+        counter shouldBe 2
```
BalmungSan: I belive `counter` should be `1`.

core/src/test/scala/neotypes/FutureAsyncSpec.scala
```diff
@@ -0,0 +1,92 @@
...
+        ex.getMessage shouldBe "Finalizer failure count: 1!" //first failure
+        resourceIsOpen shouldBe true //resource was not closed
+    }
+  }
+
+  "async future" should "return finalizer exception when finalizer fails and then succeeds" in new FutureAsyncFixture {
```
BalmungSan: Actually, the finalizer should only be called one time.

core/src/test/scala/neotypes/FutureAsyncSpec.scala
geoffjohn11: If you revert the code to the original spec and run this test you'll see this is the actual behavior.  Basically, function fa throws an exception, and in the "recoverWith" block the finalizer is executed, toggling the boolean to true.

core/src/test/scala/neotypes/FutureAsyncSpec.scala
geoffjohn11: This test will enter the flatMap block after parameter "fa" and the function "f" throws an exception when it is invoked.

core/src/test/scala/neotypes/FutureAsyncSpec.scala
geoffjohn11: I think finalizer is called once here:
```java
_ <- finalizer(a, None)
```

and then again here:

```java
case ex: Throwable =>
              finalizer(a, Some(ex))
```

BalmungSan: Yeah but it shouldn't. That is the point.

core/src/test/scala/neotypes/FutureAsyncSpec.scala
BalmungSan: Ah I see now 👍
It is a bit confusing, I think it would be better that instead of overriding values, we pass the inputs to a helper method.

core/src/test/scala/neotypes/FutureAsyncSpec.scala
geoffjohn11: If you revert the code change and run this test with original code, I think you'll find the behavior is the same

BalmungSan: Could you check that? In any case, we need to revert the changes since the actual implementation is buggy, you can see that other tests like the refined one fail.

I believe we could simplify the previous implementation using a `transformWith`, those tests would ensure that whatever change we made are secure, which is fantastic!

geoffjohn11: I wrote the test against the original functionality, and then made my code change, so I am very sure I didn't break anything.  I will revert it for the PR, however.

geoffjohn11: reverted the code change to Async

BalmungSan:
@geoffjohn11 Apart from the aesthetic changes, I made the test to really ensure what we want.

The current implementation doesn't satisfy those!
Which is great that we found thanks to this PR... but are really bad news!!!

I will try to fix it, but first I am going to try to reduce the current boilerplate and add those tests for all implementations of Async. So, if in the meantime you want to give it a look at how to fix the current implementation, that would be great.

geoffjohn11:
> @geoffjohn11 Apart from the aesthetic changes, I made the test to really ensure what we want.
> 
> The current implementation doesn't satisfy those!
> Which is great that we found thanks to this PR... but are really bad news!!!
> 
> I will try to fix it, but first I am going to try to reduce the current boilerplate and add those tests for all implementations of Async. So, if in the meantime you want to give it a look at how to fix the current implementation, that would be great.

Sure, I'll try to fix the implementation for Future

BalmungSan:
@geoffjohn11 actually the original spec wasn't wrong, but I got confused.
I just fixed that and added some other cases that were missing.

I also make the spec generic so it can work with any effect, in the following commit I will implement it for every other effect.

Nevertheless, the current implementation for future is still wrong _(but only for one case and not for almost all as the preivous commit)_.
But, I am not sure if it could be fixed...

BalmungSan:
@geoffjohn11 I have just fixed it! 🎉 

I believe this PR is ready to be merged, but I would like if you can review my changes.
Also, if @dimafeng can give it a quick look, that would be great.

core/src/test/scala/neotypes/AsyncGuaranteeSpec.scala
```diff
@@ -0,0 +1,124 @@
...
+
+    res.right.value shouldBe expectedResult
+    assertFinalizerWasCalledOnlyOnce
+  }
+
+  it should "not execute finalizer and return input exception when input fails" in new AsyncGuaranteeFixture {
+    val inputEx = new Exception("Input failed")
+
+    val ex = run(inputEx = Some(inputEx), result = Right("result"))
+
```
geoffjohn11: the run method is defined with parameter "result" first and "inputEx" second.  It's just stylistic, but I would keep the order the same as the method definition, so val ex = run( result = Right("result"), inputEx = Some(inputEx))

core/src/test/scala/neotypes/AsyncGuaranteeSpec.scala
```diff
@@ -0,0 +1,124 @@
...
+import neotypes.internal.syntax.async._
+
+import org.scalatest.{EitherValues, FlatSpec, Matchers}
+import org.scalatest.compatible.Assertion
+
+/** Base class for testing the Async[F].guarantee method. */
+abstract class AsyncGuaranteeSpec[F[_]] extends FlatSpec with Matchers with EitherValues {
+  def fToEither[T](f: F[T]): Either[Throwable, T]
+  implicit def F: Async[F]
+
```
geoffjohn11: This trait is pretty cool!

core/src/main/scala/neotypes/Async.scala
```diff
@@ -57,18 +57,9 @@ object Async {
...
-          result.recoverWith {
-            case ex: Throwable =>
-              finalizer(a, Some(ex))
-                .flatMap(_ => failed[B](ex))
-                .recoverWith {
-                  case _ => failed[B](ex)
-                }
+          f(a).transformWith {
+            case Success(b)  => finalizer(a, None).map(_ => b)
+            case Failure(ex) => finalizer(a, Some(ex)).transform(_ => Failure(ex))
```
geoffjohn11: This looks good to me.

core/src/test/scala/neotypes/AsyncGuaranteeSpec.scala
BalmungSan: @geoffjohn11 Yeah, I thought about that, and I end up making the changes like this. Because I like the idea of `input` -> `use` -> `close`. In the original method, it has to have the other order because `result` doesn't have a default value, so it has to go firts.

PR #89, opened 2020-02-25
Using property based testing

### LINKED ISSUE #47: Bug: AkkaStreamSpec randomly failing in Travis
From a while ago, the Akka Streams test suite is randomly failing in **Travis**.
I haven't be able to reproduce the error on local. Thus, I tend to assume the problem is related to the running environment, but this is not confirmed right now.

BalmungSan:
This also seems to fix #47 

@sloshy @geoffjohn11 I would also like to get your feedback.


dimafeng:
that's awesome that you managed to fix #47 🚀

core/src/test/scala/neotypes/TransactIntegrationSpec.scala
```diff
@@ -1,33 +1,63 @@
...
-      results => assert(results == expectedResults)
-    }
+  implicit def F: Async[F]
 
-  final def ensureRollbackedTransaction[E <: Throwable : ClassTag]
-                                       (txF: Transaction[F] => F[Unit])
-                                       (implicit F: Async[F]): Future[Assertion] =
+  private final def ensureCommitedTransaction[T](expectedResults: T)
+                                                (txF: Transaction[F] => F[T]): Future[Assertion] =
+    fToFuture(
```
geoffjohn11: I like the changes to this trait to unify the query logic in one place to reduce the duplication!  That's nice!  One small comment on the api, is that scala.Future is a supported effect type, but seems to have more influence than then other effect types, since in the fToFuture method we provide an implementation to convert the other effect types into Future.  I understand that this is primarily due to the influence exerted by AsyncFlatSpec, since it likes a Future[Assertion] but doesn't know what to do about F[Assertion].  One alternative would be to re-implement fToFuture to something like evaluateF and the new method evaluateF would return a T instead of a Future[T].

akka-stream/src/test/scala/neotypes/akkastreams/AkkaStreamSpec.scala
```diff
@@ -2,25 +2,31 @@ package neotypes.akkastreams
...
-      .runWith(Sink.seq[Int])
-      .map {
-        names => assert(names == (0 to 10))
+  override def streamToFList[T](stream: AkkaStream[T]): Future[List[T]] =
+    stream.runWith(Sink.seq[T]).map(_.toList).flatMap { result =>
+      Future {
+        // Delaying the answer a little seems to fix:
+        // https://github.com/neotypes/neotypes/issues/47
+        Thread.sleep(1000)
+        result
```
geoffjohn11: Sometimes it's unavoidable, but Thread.sleep can cause issues down the road.  It's good to know that the issue is a timing thing though.

BalmungSan: Futures are hard to reason about. I still do not understand what exactly is the problem, but it seems that if this operation is too fast, it tries to close the transaction twice, which results in an error.

A small delay of 1 second is enough to fix the problem.
In general, I would not be happy with this kind of solution, but since this is just a unique test case, I would just leave it.

core/src/test/scala/neotypes/TransactIntegrationSpec.scala
BalmungSan: The thing is that it would make the tests slower and sometimes will not work, because they will block.
Also, this isn't really nothing outside the standard, most projects using **IO** or other effects write their tests using `usafeRunToFuture` and the `AsyncFlatSpec`.

PR #90, opened 2020-02-28
Adding tests for stage syntax

geoffjohn11:


core/src/test/scala/neotypes/FutureTransactionUnitSpec.scala
```diff
@@ -0,0 +1,43 @@
...
+import org.neo4j.driver.v1.summary.ResultSummary
+import org.scalatest.{Matchers, WordSpec}
+import org.scalatestplus.mockito.MockitoSugar
+
+import scala.concurrent.duration.Duration
+import scala.concurrent.{Await, Future}
+import scala.util.Try
+
+class FutureTransactionUnitSpec extends WordSpec with TransactionUnitSpec[Future] with Matchers with MockitoSugar {
+
```
geoffjohn11: @BalmungSan I thought this test highlights one of the differences between the Future effect type and the others, which is if the neo4j Transaction code throws a runtime exception, the implementation of F.async for future doesn't encapsulate that into a failed future, instead an exception is thrown.

BalmungSan: That shouldn't happen!
All the effect types should have a roughly equivalent behavior. If one catch some exceptions and the other one doesn't, we need to refactor the code!

geoffjohn11:
> I gave it a quick look since I do not have too much time right now.
> 
> I like what the test checks, but I do not like that the future has a different behavior, we need to fix that!
> Also, I try to avoid mocking as much as possible, it would be interesting to see how we may avoid it.
> Finally, using the default execution context on tests is a very bad idea!
> I suggest either using an **AsyncX** specs, which provides their own EC, or create an EC of just one thread like the `AsyncGuarateeSpec`.
> 
> BTW, you shouldn't need to repeat the tests in all the other classes.

I made a small change to the Async implementation of Future so that Future now behaves the same way as the other effect types, and refactored the tests since now they all have the same assertions.  Thanks for the suggestion about avoiding the default execution context.  I still need to make that change.

BalmungSan:
@geoffjohn11 I rewrote the tests as they were basically testing the `accept*` **extension methods**, so I thought it would be clearer to just do that directly without needing a `Transaction` nor _mocks_.

I would like if you could give them a look and leave your feedback.

core/src/test/scala/neotypes/internal/syntax/StageSyntaxSpec.scala
```diff
@@ -0,0 +1,224 @@
...
+  private val effectName: String = ct.runtimeClass.getCanonicalName
+
+  def fToFuture[T](f: F[T]): Future[T]
+  def F: Async[F]
+
+  import StageSyntaxSpec.CustomException
+
+  private def testAccept[T](result: => Either[Throwable, T],
+                            inputEx: Option[Throwable] = None): Future[T] =
+    fToFuture(
```
geoffjohn11, line 21: You might be able to simplify these a little.  Something like this:

```java
  def completionStage[T](inputEx: Option[Throwable], t: T): CompletableFuture[T] = {
    inputEx.fold(ifEmpty = CompletableFuture.completedFuture(t)) { ex: Throwable =>
      val completedFuture = new CompletableFuture[T]()
      completedFuture.completeExceptionally(ex)

      completedFuture
    }
  }

  private def testAccept[T](result: => Either[Throwable, T],
                            inputEx: Option[Throwable] = None): Future[T] =
    fToFuture(
      F.async[T] { cb =>
        completionStage(inputEx, ()).accept(cb)(_ => result)
      }
    )

  private def testAcceptExceptionally[T](result: => Either[Throwable, T], inputEx: Option[Throwable] = None)
                                        (recover: PartialFunction[Throwable, Either[Throwable, T]]): Future[T] =
    fToFuture(
      F.async[T] { cb =>
        completionStage(inputEx, ()).acceptExceptionally(cb)(_ => result)(recover)
      }
    )

  private def testAcceptVoid(inputEx: Option[Throwable] = None): Future[Unit] =
    fToFuture(
      F.async[Unit] { cb =>
        completionStage[Void](inputEx, None.orNull).acceptVoid(cb)
      }
    )
```

PR #117, opened 2020-04-22
IncoercibleException Message

geoffjohn11:


core/src/main/scala/neotypes/mappers.scala
```diff
@@ -278,8 +278,8 @@ object mappers {
         value match {
           case Some(v) =>
             Try(f(v)).toEither.left.map {
-              case ex: Uncoercible => IncoercibleException(ex.getLocalizedMessage, ex)
-              case ex: Throwable   => ex
+              case ex: Uncoercible => IncoercibleException(s"${ex.getLocalizedMessage} for field [$fieldName] with value [${value}]", ex)
```
BalmungSan, line 281: Use `v` instead of `value`.

core/src/main/scala/neotypes/mappers.scala
```diff
@@ -278,8 +278,8 @@ object mappers {
         value match {
           case Some(v) =>
             Try(f(v)).toEither.left.map {
-              case ex: Uncoercible => IncoercibleException(ex.getLocalizedMessage, ex)
-              case ex: Throwable   => ex
+              case ex: Uncoercible => IncoercibleException(s"${ex.getLocalizedMessage} for field [$fieldName] with value [${v}]", ex)
```
BalmungSan, line 281: Just for the sake of style and standard, I would `for field [${fieldName}]` 
_(note the curly brackets)_

---

PS: This is a really minor thing and I wouldn't stop merging this only for this _(and simply fix it myself on my next pr)_.
However, the next comment does merit asking for a fix, so why not mentioning this too :p

core/src/test/scala/neotypes/CompositeTypesSpec.scala
```diff
@@ -117,6 +117,17 @@ class CompositeTypesSpec extends BaseIntegrationSpec[Future] {
...
 
+  it should "construct an IncoercibleException message with a field name and value" in execute { s =>
+    final case class PersonIntName(name: Int, born: Int, extra: Option[Int])
+    "match (p:Person {name: 'Charlize Theron'}) return p"
+      .query[PersonIntName]
+      .single(s)
+      .transformWith{
+        case scala.util.Failure(f) => assert(f.getMessage == "Cannot coerce STRING to Java int for field [name] with value [\"Charlize Theron\"]")
+        case scala.util.Success(_) => fail("Query succeeded, but should have failed with IncoercibleException")
+      }
```
BalmungSan, line 128: This is not the correct way to test for a failed future.

Check out other tests, there is a `recoverToExceptionIf[ExClass]` which will transform the `Future[A]` into a `Future[ExClass]` if it failed with that exception or fail if not.
Then you can `map` that exception to check the message.

Note that, if you do not want to check anything about the exception and only that it was thrown you an use `recoverToSucceedIf[ExType]`.

PR #119, opened 2020-04-27
Tests refactor - reducing boilerplate

BalmungSan:
This PR reduces the amount of boilerplate on the effect and stream tests.
It also groups all the effect tests into one big suite, which makes it easier to add new specs and sure it is implemented to all effects. _(the same was done for streams, but for now we only have one spec for those)_.

Also, I took the opportunity to make most of the core specs abstract to their effect, so now they are run for all effects, not only **Future**.


dimafeng:
@BalmungSan wow! Great job! I didn't play with this PR on my local machine but it looks great based on my "light" review. Feel free to merge it

cats-effect/src/test/scala/neotypes/cats/effect/IOSuite.scala
```diff
@@ -0,0 +1,21 @@
...
+      override final def fToFuture[T](io: IO[T]): Future[T] =
+        io.unsafeToFuture()
+
+      override final val asyncInstance: Async[IO] =
+        implicitly
+    }
+}
+
+/** Execute all the effect specs using IO. */
+final class IOSuite extends EffectSuite(IOTestkit)
```
geoffjohn11, line 21: When I run a test suite the output in sbt for the suite looks like this:
```java
[info] EffectSuite$$anon$6:

[info] EffectSuite$$anon$7:

[info] EffectSuite$$anon$8:
```
It would be a bit nicer to see AsyncGuaranteeSpec, AsyncIntegrationSpec, etc...

BalmungSan: Yeah, that is annoying but after that line, you can see the behavior line.
For example, [here](https://travis-ci.org/github/neotypes/neotypes/builds/679884593#L1131):

> [info] EffectSuite$$anon$1:
> [info] Async[scala.concurrent.Future].guarantee

That anonymous class name comes from [here](https://github.com/neotypes/neotypes/pull/119/files#diff-bc4e76c29a5d67a8f6b773b289d50ea5R36-R47):

```scala
abstract class EffectSuite[F[_]](testkit: EffectTestkit[F]) extends Suites(
  new AsyncGuaranteeSpec(testkit) {},
  new AsyncIntegrationSpec(testkit) {},
  new BasicSessionSpec(testkit) {},
  new BasicTransactionSpec(testkit) {},
  new CompositeTypesSpec(testkit) {},
  new ParameterSpec(testkit) {},
  new PathSessionSpec(testkit) {},
  new QueryExecutionSpec(testkit) {},
  new StageSyntaxSpec(testkit) {},
  new TransactIntegrationSpec(testkit) {}
)
```

The problem is that if I use concrete classes, then **Scalatests** tries to run them.

AFAIK, there is no way to solve this.
But will check once again.

BalmungSan: @geoffjohn11 oh it seems that since now the specs have a constructor, **Scalatest** doesn't try to run it. So now I am replacing the abstract classes with final classes and the names are good again :smile:

core/src/test/scala/neotypes/BaseIntegrationSpec.scala
```diff
@@ -1,43 +1,57 @@
...
+      new neo4j.TransactionWork[Unit] {
+        override def execute(tx: neo4j.Transaction): Unit =
+          tx.run(query)
       }
+    )
+  }
+
+  override final def afterStart(): Unit = {
+    if (initQuery != null) {
+      runQuery(initQuery)
```
geoffjohn11, line 37: CleaningIntegrationSpec doesn't need an initQuery value, but it must inherit one from BaseIntegrationSpec.  BaseIntegrationSpec defines a cleanDb method, but it is only used in CleaningIntegrationSpec.  Could these be broken out into a different OO hierarchy, maybe CleanIntegrationSpec and DirtyIntegrationSpec?

BalmungSan: > CleaningIntegrationSpec doesn't need an initQuery value

Well, it does, an empty one.

> BaseIntegrationSpec defines a cleanDb method, but it is only used in CleaningIntegrationSpec

Yeah, I believe I can move that method into **CleaningIntegrationSpec**.

> Could these be broken out into a different OO hierarchy

That would just make things more complicated.
A **CleaningIntegrationSpec** is an **IntegrationSpec** that start all tests with an empty db.

core/src/test/scala/neotypes/EffectSuite.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  protected final def fToFuture[T](f: F[T]): Future[T] =
+    behaviour.fToFuture(f)
+
+  protected implicit final val F: Async[F] =
+    behaviour.asyncInstance
+}
+
+/** Group all the effect specs into one big suite, which can be called for each effect. */
+abstract class EffectSuite[F[_]](testkit: EffectTestkit[F]) extends Suites(
```
geoffjohn11, line 36: I like the unifying the tests for the effects in one place, and passing in the EffectTestkit[F] for the different effect types.  One question about this approach, how would I run just one of these tests?  I can forsee the use case where a developer is making changes to a specific part of the codebase, and just wants to run the test that hits that code and not the entire suite.

BalmungSan: Yeah you can simply create a `final class JustRunThis extends TheSpecYouWant(FutureTestkit)` or you can create your own suite, or you may comment the classes you do not want to run for the main suite.

In the end, you want to ensure your changes don't break anything.
So I am ok with the fact that just wanting to run a couple of specs requiring some extra work of the developer. Because, those changes will never be committed.

core/src/test/scala/neotypes/BaseIntegrationSpec.scala
BalmungSan: Ah I remember, I create `cleanDb` on the **BaseIntegrationSpec**. Because otherwise, I would have to expose the `runQuery` method, which is very powerful.
Also, someday `cleanDb` may be useful for other specs.

So I would leave this as it is.

core/src/test/scala/neotypes/EffectSuite.scala
geoffjohn11: Sounds good, makes sense.

core/src/test/scala/neotypes/BaseIntegrationSpec.scala
geoffjohn11: 👍

PR #124, opened 2020-05-04
Accumulate case class mapping errors instead of fail-fast

geoffjohn11:


core/src/main/scala/neotypes/implicits/mappers/ResultMappers.scala
```diff
@@ -4,15 +4,15 @@ package implicits.mappers
 import java.time._
 import java.util.UUID
 
-import mappers.{ResultMapper, TypeHint, ValueMapper}
-import types.Path
-
+import neotypes.exceptions.MultipleIncoercibleException
+import neotypes.mappers.{ResultMapper, TypeHint, ValueMapper}
+import neotypes.types.Path
```
BalmungSan, line 9: All over the code base, we have been using the convention of using relative imports for importing things of the same project.
We have also kept them separated from other imports to make it clear to the eyesight.

core/src/main/scala/neotypes/implicits/mappers/ResultMappers.scala
```diff
@@ -4,15 +4,15 @@ package implicits.mappers
...
 import org.neo4j.driver.internal.types.InternalTypeSystem
 import org.neo4j.driver.internal.value.IntegerValue
 import org.neo4j.driver.v1.Value
-import org.neo4j.driver.v1.types.{Entity, IsoDuration, Node, Path => NPath, Point, Relationship}
-import shapeless.{HList, HNil, LabelledGeneric, Lazy, Witness, labelled, :: => :!:}
+import org.neo4j.driver.v1.types.{Entity, IsoDuration, Node, Point, Relationship, Path => NPath}
 import shapeless.labelled.FieldType
+import shapeless.{HList, HNil, LabelledGeneric, Lazy, Witness, labelled, :: => :!:}
```
BalmungSan, line 15: It seems you have an auto-formatter which mess with the imports.

I believe the best would be to disable it for this PR, if you want, we can discuss in the gitter room if you believe we should be following another standard. And if get to an agreement on that, open a new PR formatting all the files at once.

core/src/main/scala/neotypes/implicits/mappers/ResultMappers.scala
```diff
@@ -100,17 +100,17 @@ trait ResultMappers extends ValueMappers {
                                              reprDecoder: Lazy[ResultMapper[R]],
                                              ct: ClassTag[A]): ResultMapper[A] =
     new ResultMapper[A] {
-      override def to(value: List[(String, Value)], typeHint: Option[TypeHint]): Either[Throwable, A] =
+      override def to(value: List[(String, Value)], typeHint: Option[TypeHint], errors: Seq[Throwable]): Either[Throwable, A] =
```
BalmungSan, line 103: I like to stay away from **Seq** whenever possible. I believe **List** would be the best here.

You can take a look to [this](https://gist.github.com/djspiewak/2ae2570c8856037a7738#generic-interfaces) to understand why I do not like to use generic sequences.

core/src/main/scala/neotypes/implicits/mappers/ResultMappers.scala
```diff
@@ -147,14 +147,24 @@ trait ResultMappers extends ValueMappers {
...
+
+            tail match{
+              case HNilResultMapper if errors.nonEmpty =>
+                Left(MultipleIncoercibleException(errors.toList.reverse))
+              case _ =>
+                decodedHead match{
+                  case Left(th) =>
+                    tail.to(convertedValue, typeHint, th +: errors).map(t => labelled.field[K](null.asInstanceOf[H]) :: t)
+                  case Right(v) => tail.to(convertedValue, typeHint, errors).map(t => labelled.field[K](v) :: t)
+                }
```
BalmungSan, line 159: I would format it like:

```scala
decodedHead match {
  case Left(th) =>
     tail.to(convertedValue, typeHint, th :: errors).map(t => labelled.field[K](null.asInstanceOf[H]) :: t)

  case Right(v) =>
    tail.to(convertedValue, typeHint, errors).map(t => labelled.field[K](v) :: t)
}
```

Also, both `null` and `asInstanceOf` are code smells.
There should be a better way to make it typecheck.

---

_(I will try to play with the code on an editor to understand why you did that, but that will probably take me a couple of days)_

core/src/main/scala/neotypes/mappers.scala
```diff
@@ -13,7 +13,7 @@ import scala.util.Try
 object mappers {
   @annotation.implicitNotFound("Could not find the ResultMapper for ${A}")
   trait ResultMapper[A] { self =>
-    def to(value: List[(String, Value)], typeHint: Option[TypeHint]): Either[Throwable, A]
+    def to(value: List[(String, Value)], typeHint: Option[TypeHint], errors: Seq[Throwable] = Nil): Either[Throwable, A]
```
BalmungSan, line 16: I believe that default value is not needed at all.
Apply to all the following ones.

core/src/test/scala/neotypes/CompositeTypesSpec.scala
```diff
@@ -122,14 +122,26 @@ final class CompositeTypesSpec[F[_]](testkit: EffectTestkit[F]) extends BaseInte
...
-    recoverToExceptionIf[IncoercibleException] {
+    recoverToExceptionIf[MultipleIncoercibleException] {
       executeAsFuture { s =>
         "match (p:Person {name: 'Charlize Theron'}) return p"
           .query[PersonIntName]
           .single(s)
       }
     } map { ex =>
-      assert(ex.getMessage  == "Cannot coerce STRING to Java int for field [name] with value [\"Charlize Theron\"]")
+      assert(ex.errors.map(_.getMessage)  == List("Cannot coerce STRING to Java int for field [name] with value [\"Charlize Theron\"]"))
```
BalmungSan, line 132: **Scalatest** has a trait `LoneElement` which would be ideal in this case:

```scala
ex.errors.lonElement.getMessage == "Cannot ... "
```

core/src/test/scala/neotypes/CompositeTypesSpec.scala
```diff
@@ -122,14 +122,26 @@ final class CompositeTypesSpec[F[_]](testkit: EffectTestkit[F]) extends BaseInte
...
+
+  it should "construct an MultipleIncoercibleException message with all failures" in {
+    recoverToExceptionIf[MultipleIncoercibleException] {
+      executeAsFuture { s =>
+        "match (p:Person {name: 'Charlize Theron'}) return p"
+          .query[PersonIntNameBornBool]
+          .single(s)
+      }
+    } map { ex =>
+      assert(ex.errors.map(_.getMessage)  == List("Cannot coerce STRING to Java int for field [name] with value [\"Charlize Theron\"]", "Cannot coerce INTEGER to Java boolean for field [born] with value [1975]"))
```
BalmungSan, line 144: I would split the list into multiple lines to avoid the need to scroll.

core/src/main/scala/neotypes/implicits/mappers/ResultMappers.scala
geoffjohn11: Thanks for the link!  That's a great explanation

core/src/main/scala/neotypes/implicits/mappers/ResultMappers.scala
geoffjohn11: Whoops, yeah I'll disable it.

PR #131, opened 2020-05-20
ZIO RC19-2 Fix

geoffjohn11:


zio-stream/src/main/scala/neotypes/zio/stream/ZioStreams.scala
```diff
@@ -18,13 +18,11 @@ trait ZioStreams {
...
 
       override final def onComplete[T](s: ZioStream[T])(f: => Task[Unit]): ZioStream[T] =
-        ZStream.flatten(
-          ZStream.bracket(Task(s)) { _ =>
-            f.orDie
-          }
-        )
+        ZStream.bracket(Task(s)) { _ =>
+          f.orDie
+        }.flatMap(identity)
```
BalmungSan, line 23: It would be better to call `flatten` instead of `flatMap(identity)`

zio-stream/src/main/scala/neotypes/zio/stream/ZioStreams.scala
```diff
@@ -18,13 +18,11 @@ trait ZioStreams {
...
 
       override final def fToS[T](f: Task[ZioStream[T]]): ZioStream[T] =
-        ZStream.flatten(ZStream.fromEffect(f))
+        ZStream.fromEffect(f).flatMap(identity)
```
BalmungSan, line 26: It would be better to call `flatten` instead of `flatMap(identity)`

PR #140, opened 2020-06-04
Upgrade neo4j java driver to version 4

### LINKED ISSUE #58: Upgrade to Neo4j 4
A new version fo the Neo4j Java driver has been released.

The [changelogs](https://github.com/neo4j/neo4j-java-driver/wiki/4.0-changelog) state that probably the most important change is the support of an **Rx** driver. It would be great if we can provide wrappers for those too for our **Streaming** implementations, for example since I use `http4s` I end up converting everything into an `fs2` **Stream**.

BalmungSan:
Fixes #58 

TODO list:

- [x] Upgrade the driver and the docker container.
- [x] Fix compilation errors.
- [x] Fix tests.
- [x] Fix docs.
- [x] Cover all the new API _(missing [`TransactionConfig`](https://neo4j.com/docs/api/java-driver/current/org/neo4j/driver/TransactionConfig.html))_.

I am still missing this:

- [ ] Provide wrappers for the new Rx module.

But I believe that is work for another PR.
This is ready for review.


PhilAndrew:
Looking forward to this

BalmungSan:
@dimafeng I added the `TransactionConfig` docs!

> I see that this brings some incompatibilities with the old driver - do you think we need to release it with a new versioning schema.

Yeah, this should be a major release... however, I still do think we are ready to release the `1.0.0` version _(at least not until the concurrency issues with `Session` are resolved #44)_.
So, I believe we should just release version `0.14.0`. BTW, I just added a compatibility matrix to the docs.


Anyways, I will proceed to merge this as well as the pending upgrade dependencies PRs so everything is ready for you to make the new release.

PR #163, opened 2020-07-24
Making Session thread safe

### LINKED ISSUE #44: Discussion: Session + Concurrency

BalmungSan:
Fixes #44 

Adding a simple locking mechanism which ensures only one **Transaction** per **Session** can be active at time.
For pure effect types, the block for the lock is just semantic _(meaning no real thread is blocked)_. For **Future**, I used `blocking` which will notify the EC that the next operation will block; that is a _best effort_ since the EC may decide to allocate a new thread for not blocking a compute thread, but it may also just ignore it.

------

## Edit

~I still need to update the documentation to be very explicit about how this works and made clear that when not using managed transactions one has to be very careful of just calling  `commit` or `rollback` once; otherwise, the system is left in unspecified behavior _(most probably a failure or a deadlock)_.~

Done.

### DIFF
--- a/core/src/main/scala/neotypes/Async.scala
+++ b/core/src/main/scala/neotypes/Async.scala
@@ -8,32 +8,32 @@ import scala.util.{Failure, Success}
 
 @annotation.implicitNotFound("The effect type ${F} is not supported by neotypes")
 trait Async[F[_]] {
-  type R[A]
+  private[neotypes] type R[A]
 
-  trait Lock {
+  private[neotypes] trait Lock {
     def acquire: F[Unit]
     def release: F[Unit]
   }
 
-  def async[A](cb: (Either[Throwable, A] => Unit) => Unit): F[A]
+  private[neotypes] def async[A](cb: (Either[Throwable, A] => Unit) => Unit): F[A]
 
-  def delay[A](a: => A): F[A]
+  private[neotypes] def delay[A](a: => A): F[A]
 
-  def failed[A](e: Throwable): F[A]
+  private[neotypes] def failed[A](e: Throwable): F[A]
 
-  def flatMap[A, B](fa: F[A])(f: A => F[B]): F[B]
+  private[neotypes] def flatMap[A, B](fa: F[A])(f: A => F[B]): F[B]
 
-  def guarantee[A, B](fa: F[A])
-                     (f: A => F[B])
-                     (finalizer: (A, Option[Throwable]) => F[Unit]): F[B]
+  private[neotypes] def guarantee[A, B](fa: F[A])
+                                       (f: A => F[B])
+                                       (finalizer: (A, Option[Throwable]) => F[Unit]): F[B]
 
-  def makeLock: F[Lock]
+  private[neotypes] def makeLock: F[Lock]
 
-  def map[A, B](fa: F[A])(f: A => B): F[B]
+  private[neotypes] def map[A, B](fa: F[A])(f: A => B): F[B]
 
-  def recoverWith[A, B >: A](fa: F[A])(f: PartialFunction[Throwable, F[B]]): F[B]
+  private[neotypes] def recoverWith[A, B >: A](fa: F[A])(f: PartialFunction[Throwable, F[B]]): F[B]
 
-  def resource[A](input: F[A])(close: A => F[Unit]): R[A]
+  private[neotypes] def resource[A](input: F[A])(close: A => F[Unit]): R[A]
 }
 
 object Async {
--- a/core/src/main/scala/neotypes/Async.scala
+++ b/core/src/main/scala/neotypes/Async.scala
@@ -3,7 +3,7 @@ package neotypes
 import java.util.concurrent.ArrayBlockingQueue
 
 import scala.concurrent.{Await, ExecutionContext, Future, Promise, blocking}
-import scala.concurrent.duration.Duration
+import scala.concurrent.duration._ // Provides the second extension method.
 import scala.util.{Failure, Success}
 
 @annotation.implicitNotFound("The effect type ${F} is not supported by neotypes")
@@ -93,6 +93,6 @@ object Async {
         fa.recoverWith(f)
 
       override final def resource[A](input: Future[A])(close: A => Future[Unit]): A =
-        Await.result(input, Duration.Inf)
+        Await.result(input, 1.second)
     }
 }
--- a/core/src/main/scala/neotypes/Async.scala
+++ b/core/src/main/scala/neotypes/Async.scala
@@ -1,11 +1,19 @@
 package neotypes
 
-import scala.concurrent.{ExecutionContext, Future, Promise}
+import java.util.concurrent.ArrayBlockingQueue
+
+import scala.concurrent.{Await, ExecutionContext, Future, Promise, blocking}
+import scala.concurrent.duration.Duration
 import scala.util.{Failure, Success}
 
 @annotation.implicitNotFound("The effect type ${F} is not supported by neotypes")
 trait Async[F[_]] {
   type R[A]
+
+  trait Lock {
+    def acquire: F[Unit]
+    def release: F[Unit]
+  }
 
   def async[A](cb: (Either[Throwable, A] => Unit) => Unit): F[A]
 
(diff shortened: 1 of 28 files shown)


core/src/main/scala/neotypes/Transaction.scala
```diff
@@ -14,105 +15,140 @@ import org.neo4j.driver.exceptions.NoSuchRecordException
...
+            .accept(cb) { statementResultCursor =>
+              Right(S.init(() => nextAsyncToF(statementResultCursor.nextAsync())))
+            }
+        }
+      )
+
+    override final def commit: F[Unit] =
+      F.async[Unit] { cb =>
+        transaction.commitAsync().acceptVoid(cb)
+      } flatMap { _=>
```
dimafeng, line 143: If `transaction.commitAsync().acceptVoid(cb)` fails, we still want to release the lock, don't we?

site/src/main/mdoc/driver_session_transaction.md
```diff
@@ -13,9 +13,18 @@ A `Driver` is basically the connection with the Database. Usually, you would onl
 A `Session` provides a context for performing operations _(`Transactions`)_ over the database. You may need as many as concurrent operations you want to have.
 A `Transaction` is a logical container for an atomic unit of work. Only one transaction may exist in a `Session` at any point in time.
 
-## Transaction managment
+## Session thread safety
 
-Each `Transaction` has to be started, used and finally either, committed or rolled back.
+Unlike its **Java** counterpart, `neotypes.Session` is thread safe.
+We achieve that by using a simple locking mechanism that ensures only one `Transaction` can be created `Session`.
```
dimafeng, line 19:
```
only one `Transaction` can be created `Session`
```
Should it be `one one Transaction at a time for each Session` or something like that

core/src/main/scala/neotypes/Async.scala
```diff
@@ -64,13 +73,26 @@ object Async {
...
       override final def map[A, B](fa: Future[A])(f: A => B): Future[B] =
         fa.map(f)
 
       override final def recoverWith[A, B >: A](fa: Future[A])(f: PartialFunction[Throwable, Future[B]]): Future[B] =
         fa.recoverWith(f)
 
-      override final def resource[A](input: => A)(close: A => Future[Unit]): A =
-        input
+      override final def resource[A](input: Future[A])(close: A => Future[Unit]): A =
+        Await.result(input, Duration.Inf)
```
dimafeng, line 96: Maybe we can make duration configurable somehow? I guess if there is a logical error in the user code that prevents the lock from releasing, user would expect it to time out eventually.

site/src/main/mdoc/driver_session_transaction.md
```diff
@@ -13,9 +13,18 @@ A `Driver` is basically the connection with the Database. Usually, you would onl
...
+If you want / need to run multiple queries concurrently then you need to create multiple `Sessions` and load-balance the petitions across them yourself.
```
dimafeng, line 20: - I'm not sure I don't understand what `petitions` means in the current context. Did you mean something else?

- Pick one 😁 
```
`Sessions`
``` 
-> 
```
`Session`s
```

core/src/main/scala/neotypes/Async.scala
```diff
@@ -64,13 +73,26 @@ object Async {
           }
         }
 
+      override final def makeLock: Future[Lock] =
+        Future.successful {
+          val q = new ArrayBlockingQueue[Unit](1)
```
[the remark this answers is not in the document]
BalmungSan: As always, the future implementation was the most difficult one.

It literary took me 15 minutes to implement all the other effects, whereas this one took me like 5 hours.

Between all the things I tried, **ReentrantLocks** were the first _(that is why I named it `Lock`)_.
They won't work because the EC may schedule two futures in the same thread and it will fail _(it happened multiple times when running tests)_.

core/src/main/scala/neotypes/Transaction.scala
BalmungSan: You are goddamn right!
We need to use `guarantee` here.

core/src/main/scala/neotypes/Async.scala
BalmungSan: There are only two uses of `resource` in all the codebase.
The creation of a **Driver** and the creation of a **Session**, neither of them are waiting for a look and both should be already complete when calling this, so I just used `Await` + `Duration.Inf` to get the plain value from the futures.

dimafeng: Gotcha! I hadn't realized that

BalmungSan:
Thanks for the review @dimafeng 

I have pushed a new commit addressing your comments, let me know if I missed something.

core/src/main/scala/neotypes/Async.scala
geoffjohn11: Could also
```java
input.value.get.get
```

BalmungSan: @geoffjohn11 it should be there, but it could have a couple of milliseconds of delay _(due to the threading allocation of future)_.
Also, all those three methods are unsafe; so I prefer to use `Await` to ensure the value is correctly retrieved.

Now, if we all agree we can use something like `1.second` or `10.seconds` to ensure this doesn't block forever.

core/src/main/scala/neotypes/Async.scala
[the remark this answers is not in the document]
geoffjohn11: Is there any concern about not having lock fairness?

core/src/main/scala/neotypes/Async.scala
dimafeng: 1.second or 10.seconds sounds good to me

core/src/main/scala/neotypes/Session.scala
```diff
@@ -6,28 +6,47 @@ import internal.syntax.stage._
...
-    transaction(NeoTransactionConfig.empty)
-
-  def transaction(config: NeoTransactionConfig)(implicit F: Async[F]): F[Transaction[F]] =
-    F.async { cb =>
-      session.beginTransactionAsync(config).accept(cb) { tx =>
-        Right(new Transaction(tx))
+sealed trait Session[F[_]] {
+  def transaction: F[Transaction[F]]
+
+  def transaction(config: NeoTransactionConfig): F[Transaction[F]]
```
geoffjohn11, line 12: I'm a bit curious about the shift to remove the implicit Async[F].  Does this represent a cleaner API?

core/src/test/scala/neotypes/ConcurrentSessionSpec.scala
```diff
@@ -1,24 +1,32 @@
...
+import neotypes.internal.syntax.async._
 import neotypes.implicits.mappers.all._
 import neotypes.implicits.syntax.cypher._
 import org.neo4j.driver.exceptions.ClientException
+import org.scalatest.matchers.should.Matchers
+import scala.concurrent.ExecutionContext
 
 /** Base class for testing the concurrent use of a session. */
-final class ConcurrentSessionSpec[F[_]](testkit: EffectTestkit[F]) extends BaseIntegrationSpec[F](testkit) {
+final class ConcurrentSessionSpec[F[_]](testkit: EffectTestkit[F]) extends BaseIntegrationSpec[F](testkit) with Matchers {
```
geoffjohn11, line 11: Nice to see this test working properly.

core/src/main/scala/neotypes/Session.scala
BalmungSan: Yeah, basically the idea that users of **Session** shouldn't need the `Async` in scope, just the creator.
This also means that it may be possible to implement more **Sessions** that require different constraints.

PR #164, opened 2020-08-02
Add wrappers for the new Rx module 

### LINKED ISSUE #149: Provide wrappers for the new Rx module
Since #140 we are now on version `4` of the driver, which provides new RX based sessions and transactions.
Those should be mapped to any of our streaming abstractions.

BalmungSan:
Fixes #149 

This PR aims to provide two new wrappers for the new Rx modules of the official **neo4j** driver.
I decided to model them as extensions of their Async counterparts. This way, users do not need to manage both, they just need to decide if they want streaming capabilities or not when instantiating their `Session`s.

Also, my idea is to still keep the old `stream` capability but deprecate it.
And in a future release replace the old `stream` methods with the `streamRx` variants introduced here and deprecate the `*Rx` names so those are removed on an even later release.
Do we all agree on this? Or someone has a different opinion / idea?

------

_This will be a big PR and I am still missing a lot, so I am opening this for early feedback_.

------

### Roadmap

- [x] Core implementation.
- [x] Add tests.
- [x] Implement the new Stream abstraction for all stream modules.
- [x] ~Add deprecations~ Remove old API.
- [x] Updated docs.


core/src/main/scala/neotypes/Stream.scala
```diff
@@ -1,14 +1,40 @@
...
+  // Legacy module ------------------------------------------------------------
+  private[neotypes] def init[A](value: () => F[Option[A]]): S[A]
+
+  private[neotypes] def onComplete[A](s: S[A])(f: => F[Unit]): S[A]
+
+  private[neotypes] def fToS[A](f: F[S[A]]): S[A]
+  // --------------------------------------------------------------------------
+
+  // New (Rx) module ----------------------------------------------------------
+  private[neotypes] def fromRx[A](publisher: Publisher[A]): S[A]
```
geoffjohn11, line 20: Are you leaking the reactor-core Publisher class into the API?  Would it be better to do something similar to the CompletionStageOps which keeps the java CompletionStage out of the API?

geoffjohn11: My thought is to do something like this (exception case is missing)
```java
  implicit class PublisherStageOps[A](private val underlying: Publisher[A]) extends AnyVal {
    private final def acceptImpl[B](cb: Either[Throwable, B] => Unit)
                                   (f: A => Either[Throwable, B])
                                   (g: Throwable => Either[Throwable, B]): Unit = {
      internal.utils.void(
        Flux.from(underlying)
          .doOnError(th => cb(g(th)))
          .doOnNext(a => cb(f(a)))
      )
    }

    def accept[B](cb: Either[Throwable, B] => Unit)
                 (f: A => Either[Throwable, B]): Unit =
      acceptImpl(cb)(f)(defaultExHandler)

    def acceptExceptionally[B](cb: Either[Throwable, B] => Unit)
                              (f: A => Either[Throwable, B])
                              (g: PartialFunction[Throwable, Either[Throwable, B]]): Unit =
      acceptImpl(cb)(f)(g.orElse(defaultExHandler))

    def acceptVoid(cb: Either[Throwable, Unit] => Unit)
                  (implicit ev: A =:= Void): Unit = {
      internal.utils.void(ev)
      acceptImpl(cb)(_ => Right(()))(defaultExHandler)
    }
  }
```

BalmungSan: @geoffjohn11 technically speaking neither **Async** nor **Stream** are public API, they are implementation details _(that is why I made all their methods private to the `neotypes` package)_.
Users only need to ensure there is an implicit instance in scope for their chosen effect and stream types.

For **Async** it was good to only require a callback because it was more flexible. And, in any case, they do not provide direct support for `CompletionStage` but rather plain callbacks.

However, a reactive stream is a very different story. Your implementation is eager, it will only work for something like `AkkaStreams`, it won't work for `fs2` for example.
I know that all streaming libraries we want to support provide a dedicated and optimized way to transform a `Publisher` into their corresponding implementations.
And those are very [complex](https://github.com/functional-streams-for-scala/fs2/blob/main/reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala) and trying to just reimplement them would be a waste.

geoffjohn11: @BalmungSan Thanks for the reply!

core/src/main/scala/neotypes/Driver.scala
```diff
@@ -78,7 +73,45 @@ final class Driver[F[_]] private[neotypes] (private val driver: NeoDriver) {
...
+    * @param config the [[SessionConfig]] used to initialize the session.
+    * @param F asynchronous effect type.
+    * @tparam S streaming type dependant on the effect type F.
+    * @return a single element streaming with the [[StreamingSession]].
+    */
+  def streamingSession[S[_]](config: SessionConfig)
+                            (implicit F: Async[F], S: Stream.Aux[S, F]): S[StreamingSession[S, F]] = {
+    val session = F.makeLock.flatMap { lock =>
+      F.delay {
+        Session(F, S, driver.rxSession(config))(lock)
```
geoffjohn11, line 108: Should "Session" be differentiated by calling it something to call out it is reactive, like "RxSession"?

core/src/main/scala/neotypes/Stream.scala
geoffjohn11, line 20: Should Publisher[A] be wrapped in effect type F?

BalmungSan: I once had the same doubt. But, a `Publisher` is a lazy _(and somewhat pure)_ value.
One only has to ensure that it was not already subscribed by someone else.

BTW, if you look how to convert a `Publisher` into any of our streams you will see that neither requires it being wrapped in `F`.
Take a look to [this](https://github.com/BalmungSan/scala-mongodb-streaming).

core/src/main/scala/neotypes/Driver.scala
BalmungSan: So we have two session classes so we need two factories.
We have three options.

+ Overload the previous factory _(`Session.apply`)_ - this is the one I choose.
+ Use two different methods like `Session.async` & `Session.streaming` _(or `Session.rx`)_ - this is the one I used for `Driver.streamingSession` since overloading won't work and being an user-facing API I wanted it to be more clear.
+ Keep using `apply` but in the companion object of `StreamingSession` - actually maybe this one would have been the best option?

I really do not have any strong reasons for one or the other if you and @dimafeng prefer a different one I would be happy to change them.

---

BTW, the same applies to **Transaction**.

core/src/test/scala/neotypes/BaseIntegrationSpec.scala
```diff
@@ -21,53 +20,55 @@ abstract class BaseIntegrationSpec[F[_]](testkit: EffectTestkit[F]) extends Base
       .configure(_.addEnv("NEO4JLABS_PLUGINS", "[\"graph-data-science\"]"))
       .configure(_.withImagePullPolicy(PullPolicy.ageBased(toJavaDuration(1.day))))
 
-  protected lazy final val driver =
+  protected final lazy val driver =
     neo4j.GraphDatabase.driver(container.boltUrl)
 
-  private lazy final val neo4jSession =
-    driver.session()
```
BalmungSan, line 28: Keeping two sessions was causing troubles, so I had to remove it.
Anyways, this change allows us to switch between the async and the streaming session so we can ensure the new wrappers are totally backwards compatible.

core/src/test/scala/neotypes/CleaningIntegrationSpec.scala
```diff
@@ -7,13 +7,14 @@ import scala.concurrent.{Await, Future}
...
-      super.withFixture(test)
-    } lastly {
-      this.cleanDb()
-    }
+    val r = for {
+      o <- super.withFixture(test).toFuture
+      _ <- this.cleanDB()
+    } yield o
+
+    new FutureOutcome(r)
```
BalmungSan, line 17: This has the consequence that if a test fails then the clean won't be executed which probably will result in more test fails, but I was unable to keep the previous blocking clean without breaking **Scalatest**.

BalmungSan:
Took me way more than what I would have wanted but I officially finished this!

Now, the only thing missing would be fixing any review comments that you may have.
@geoffjohn11 @dimafeng

dimafeng (approved):
Looks great! 👍

akka-stream/src/main/scala/neotypes/akkastreams/AkkaStreams.scala
```diff
@@ -15,49 +19,71 @@ trait AkkaStreams {
     * @param ec implicit scala.concurrent ExecutionContext that will be passed down to all Akka Streams operations
```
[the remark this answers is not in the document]
BalmungSan: Ups 😅 

I currently trying to reproduce and fix the deadlock that is happening with **AkkaStreams** on **Travis**, so I will change this with that.

BalmungSan:
Due to the discussion on https://github.com/neo4j/neo4j-java-driver/issues/797 we need to change the way we use **Sessions** which brings a whole new way of implementing this change _(as well as #178)_.
I will start from scratch with both changes today and hopefully, I will have at least a draft for the weekend.

PR #180, opened 2020-09-26
Adding ReadOnlySession

### LINKED ISSUE #178: Add support for read only transactions
The neo4j java driver provides a method on AsyncSession to allow a user to specify that a transaction should not mutate data on the neo4j server: [ReadTransactionAsync](https://neo4j.com/docs/driver-manual/current/session-api/asynchronous/). This allows optimisation in clustered environments - such as routing read-only queries to read replicas.

It should be noted that introspection does not occur on the neo4j driver side, and it does not check to see that a query won't mutate data before routing the request, but the read replica will reject it.

geoffjohn11:
Closes #178

### DIFF
--- a/core/src/main/scala/neotypes/Session.scala
+++ b/core/src/main/scala/neotypes/Session.scala
@@ -4,6 +4,7 @@ import java.util.concurrent.{CompletableFuture, CompletionStage}
 
 import internal.syntax.async._
 import internal.syntax.stage._
+
 import org.neo4j.driver.{TransactionConfig => NeoTransactionConfig}
 import org.neo4j.driver.async.{AsyncTransaction, AsyncTransactionWork, AsyncSession => NeoAsyncSession}
 
@@ -56,11 +57,9 @@ object Session {
       lock.acquire.flatMap { _ =>
         F.async { cb =>
           session.readTransactionAsync(new AsyncTransactionWork[CompletionStage[T]] {
-            override def execute(tx: AsyncTransaction): CompletionStage[T] = {
+            override def execute(tx: AsyncTransaction) = {
               val completableF = new CompletableFuture[T]()
-              txf(Transaction(F, tx)(lock)).map {
-                completableF.complete
-              }.recover{
+              txf(Transaction(F, tx)(lock)).map(completableF.complete).recover{
                 case th: Throwable => completableF.completeExceptionally(th)
               }
               completableF
--- a/core/src/main/scala/neotypes/Session.scala
+++ b/core/src/main/scala/neotypes/Session.scala
@@ -53,8 +53,8 @@ object Session {
     override final def readOnlyTransact[T](txf: Transaction[F] => F[T]): F[T] =
       readOnlyTransact(NeoTransactionConfig.empty())(txf)
 
-    override final def readOnlyTransact[T](config: NeoTransactionConfig)(txf: Transaction[F] => F[T]): F[T] =
-      lock.acquire.flatMap { _ =>
+    override final def readOnlyTransact[T](config: NeoTransactionConfig)(txf: Transaction[F] => F[T]): F[T] = {
+      val fOfT: F[T] = lock.acquire.flatMap { _ =>
         F.async { cb =>
           session.readTransactionAsync(new AsyncTransactionWork[CompletionStage[T]] {
             override def execute(tx: AsyncTransaction) = {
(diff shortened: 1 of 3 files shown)


geoffjohn11:
Need to add tests

BalmungSan (approved):
:+1:

Now we need to discuss and decide how we will use this need `readOnlyTransact`.
I see three options:

1. Use the read-only transact under the hood of all the read operations like single and list, and if someone wants to write something during a list, it has to be done using `Session.transact`.
**Pros**: Clean API, less work and it could improve the performance of some users.
**Cons**: We may end up breaking other people code and force them to change the API they use.

2. Leave the code as it is, so if someone wants to use a read-only transaction it has to be done explicitly using `Session.readOnlyTransact`.
**Pros**: We can be sure we are not breaking anything and code using read-only transactions is explicit about it.
**Cons**: Assuming most people are already doing read-only operations when using things like `single` and `list`, they would need to change the code to take advantage of this.

3. Add variants like `readOnlySingle`.
**Pros**: We can be sure we are not breaking anything and code using read-only transactions is explicit about it. Also, it may be said that this option provides a more standard API.
**Cons**: The same as the previous one, plus more code to maintain.

@geoffjohn11 as the author of the change, which one do you like the most?
@dimafeng what do you think?
@bengraygh you were the one that wanted this feature, what are your thoughts on this?
I will be honest that I would like option 1, IMHO one shouldn't be doing writing inside a reading operation, so for me, it makes sense to make that use case more explicit.

-----

BTW, another important discussion about this change is who will integrate this with #164 
I believe #164 is already big enough so I would prefer if this waits until we merge that one and then in this PR do the integration, but not sure what @geoffjohn11 thinks about that.

geoffjohn11:
> @geoffjohn11 as the author of the change, which one do you like the most?

I think option 1 is my preference.

> BTW, another important discussion about this change is who will integrate this with #164
> I believe #164 is already big enough so I would prefer if this waits until we merge that one and then in this PR do the integration, but not sure what @geoffjohn11 thinks about that.

Sure, this change can wait.

dimafeng:
I like option 1 too.

bengraygh:
Bit late to the party, but I agree that option 1 is the right choice here.

BalmungSan:
@geoffjohn11 hope you didn't mind me hijacking this PR I was reviewing your last change and I was about to propose that you should just use `guarantee`
But then I realized the previous implementation has a problem; except for **Futures** that are eager, the inner `F` you created inside `work`  will never be executed; I just fixed that as well as some failing tests.

After #164 is merged this PR will have a fun time solving conflicts and implementing `readOnlyTransact` for the **StreamingSession** 😬

core/src/main/scala/neotypes/Session.scala
```diff
@@ -44,7 +50,30 @@ object Session {
...
+      lock.acquire.flatMap { _ =>
+        F.async[F[T]] { cb =>
+          val work: AsyncTransactionWork[CompletionStage[Unit]] = (tx: NeoAsyncTransaction) => {
+            val cf = new CompletableFuture[Unit]()
+
+            cb(Right(
+              F.delay(Transaction(F, tx)(lock)).guarantee(txf) {
+                case (tx, None)    => tx.commit.flatMap(_ => F.delay(internal.utils.void(cf.complete(()))))
+                case (tx, Some(_)) => tx.rollback.flatMap(_ => F.delay(internal.utils.void(cf.complete(()))))
+              }
```
BalmungSan, line 66: #164 will introduce a helper function called `txFinalizer` which we should use here like `.guarantee(txf)(txFinalizer.andThen(_.flatMap(_ => F.delay(internal.utils.void(cf.complete(()))))))`

geoffjohn11:
> But then I realized the previous implementation has a problem; except for **Futures** that are eager, the inner `F` you created inside `work` will never be executed; I just fixed that as well as some failing tests.

Apologies for this mistake @BalmungSan !  I took a look cats-effect and the implementation of guarantee is calls `.use(f)` which says:  Allocates a resource and supplies it to the given function.

This is the reason why `guarantee` evaluates the F.delay effect?

BalmungSan:
@geoffjohn11 

> Apologies for this mistake @BalmungSan

No worries, I reviewed the change like three times and never noticed it. Mixing pure and impure code is hard!

> This is the reason why guarantee evaluates the F.delay effect?

Not really, if you look what I made was an `Async` that returned an **F** and then I flattened it.
This is one of the wonders of FP, computation is a value so I just returned it into the effect system to execute it in the appropriate moment.

BalmungSan:
Sadly also closed due #221 because of https://github.com/neo4j/neo4j-java-driver/issues/797

PR #199, opened 2020-11-09
WIP: Add auto and semiauto derivation

### LINKED ISSUE #197: Shapeless derivation ignores instances in companion object
## Background

```scala
import neotypes.implicits.mappers.results._

final case class ObjectScope(value: String)

object ObjectScope {
  val const = ObjectScope("const")
  implicit val resultMapper: ResultMapper[ObjectScope] = ResultMapper.const(const)
}

ResultMapper[ObjectScope].to(Nil, None) // Left(neotypes.exceptions$PropertyNotFoundException: Property value not found)

// Expected Right(ObjectScope(const))
```

`ResultMapper[ObjectScope]` expands into 
```ResultMapper[ObjectScope](neotypes.implicits.mappers.results.ccMarshallable)```.

From the compiler perspective, everything is correct. From the user perspective, it's a painful experience.  

Instances for alias/opaque/newtype types are ignored.

## Workaround

```
import neotypes.implicits.mappers.results.{ccMarshallable => _, _}
```

## Solution

What if we follow the auto/semiauto pattern here? 

1) Move shapeless-based derivation into a separate folder `neotypes.generic`
2) Introduce `auto` and `semiauto` derivation
3) Use dedicated types for the shapeless-specific result mappers: `DerviedResultMapper` for product and `RepResultMapper` for HList.

## Pros/Cons

Pros:
- `neotypes.generic.semiauto` can work together with https://github.com/scalaz/scalaz-deriving
- primitive instances are isolated
- more control about imports
- more options for external instances: manual, magnolia-based, etc

Cons:
- Wildcard import of `neotypes.generic.auto._` still leads to the problem explained in the post. But at least instances for primitives can be imported safely `import neotypes.implicits.mappers.results._`
- the change is source/binary breaking

What's your opinion on this topic? If you are interesting, I can pick it up and contribute.

### PR DESCRIPTION
**Pull request by iRevive**
In the scope of the PR, I moved shapeless-based derivation to `neotypes.generic.auto` and `neotypes.generic.semiauto`.

Closes #197

### CODE REVIEW

File: core/src/test/scala/neotypes/generic/auto/ResultMapperAutoSpec.scala [hunk 1]
```diff
@@ -0,0 +1,157 @@
...
+      val input = List(
+        ("string", new StringValue("twelve chars")),
+        ("int", new IntegerValue(12))
+      )
+
+      val result = mapper.to(input, None)
+      assert(result == Right(("twelve chars", 12)))
+    }
+
+    "should prioritize an instance from companion object over derived" in {
```
iRevive, line 56: An instance from the companion object has the highest priority

core/src/main/scala/neotypes/mappers.scala
```diff
@@ -148,6 +148,11 @@ object mappers {
       }
   }
 
+  private[mappers] sealed trait ResultMapperLowPriority {
```
iRevive, line 151: This change brings the correct prioritization of the typeclasses

BalmungSan (commented):
Hi @iRevive 

First of all, thanks for this change! _(as well as for removing all those unused imports, sorry about that)_

In general, I really like this change _(especially that you also changed the docs!)_ and for what I can see this follows what **circe** does so it seems like the correct way to do this _(I have to be honest that I really do not have too much experience with those things)_

I have just three comments:

+ Minor style things like: some files do not have an `EOL` before the `EOF`, the order of imports _(we have 3 blocks relative imports form the project itself, imports from external libraries, imports from **Java** and **Scala**)_, and some spacing. But being honest, we should just have some standard **scalafmt** config; so no worries about this _(feel free to change that or leave it as it is right now)_
+ What is the reason for having `hlist` and `product` sub-packages? I would just have a single `generic` package with an `auto` and `semiauto` objects inside it; like **circe**.
+ Maybe we should move all the primitive instances into the companion object of `ValueMapper` and adding an `implicit def fromValueMapper` into the companion object of `ResultMapper` so one only needs to either import `auto._` or call `deriveDecoder` and that is all _(the same should be done for `ParameterMapper`)_. What do you think? And if you also like the idea, would you like contributing also that in this or in another PR?

@dimafeng was the one that made all the **Shapeless** stuff. What do you think about this change?

iRevive:
@BalmungSan thank you for the review.

> 1. EOL/EOF, imports, styling

Added empty line before the end of the files. Adjusted the import order.  
I've updated the formatting of the method (similarly to what it was looking before).

> 2. I would just have a single generic package with an auto and semiauto objects inside it; like circe.

You are right, it should be that way. I was experimenting with different derivations (auto/tuple/hlist, etc) but it does not make sense to split it up. 

I've moved all methods into `semi` and `semiauto` objects.

> 3. Maybe we should move all the primitive instances into the companion object of ValueMapper

I like this one. I will experiment with it in the current branch.

iRevive:
> implicit def fromValueMapper into the companion object of ResultMapper

Unfortunately, it's not that smooth as expected. When ResultMapper has an implicit conversion from ValueMapper, it leads to a recursive loop between `ResultMapper.fromValueMapper` and `ValueMapper.ccValueMarshallable`.

Therefore I kept instance forwarding like it was before:
```scala
implicit final val BooleanResultMapper: ResultMapper[Boolean] =
  ResultMapper.fromValueMapper
```

core/src/test/scala/neotypes/AlgorithmSpec.scala
```diff
@@ -1,8 +1,10 @@
 package neotypes
 
+import neotypes.generic.auto._
 import neotypes.implicits.mappers.all._
 import neotypes.implicits.syntax.string._
 import neotypes.internal.syntax.async._
+import neotypes.mappers.ResultMapper
```
BalmungSan, line 7: Do we really need this import?

site/src/main/mdoc/overview.md
```diff
@@ -89,11 +89,13 @@ If you need to support your return types for this type of queries, you can provi
 * `set(session)` - runs a query and returns a **Set** of results.
 * `vector(session)` - runs a query and returns a **Vector** of results.
 * `map(session)` - runs a query and returns a **Map** of results _(only if the elements are tuples)_.
-* `collectAs(Col)(session)` - runs a query and retunrs a **Col** of results _(where **Col** is any kind of collection)_.
+* `collectAs(Col)(session)` - runs a query and returns a **Col** of results _(where **Col** is any kind of collection)_. Might require the following import `import neotypes.implicits.mappers.collections._`.
```
BalmungSan, line 92: I would be more explicit. like: _"If you are in `2.12` or you are cross-compiling with `2.12` you need to `import neotypes.implicits.mappers.collections._` or you can import `scala.collection.compat._`"_

BalmungSan (approved):
LGTM! :+1:

site/src/main/mdoc/overview.md
iRevive: Done

core/src/test/scala/neotypes/AlgorithmSpec.scala
iRevive: Good catch. 
Removed

BalmungSan (approved):

dimafeng:
@iRevive @BalmungSan looks good to me too 👍

PR #207, opened 2020-11-18
Github actions

dimafeng:
WIP github actions
- [x] regular build
- [x] microsite build
- [x] disable travis
- [x] mergify rules


BalmungSan:
BTW, remember to also update the **mergify** rules.

dimafeng:
@BalmungSan do you know anything about unidoc? I keep getting this:
```
[debug] Done updating root
[error] java.util.NoSuchElementException: None.get
[error] 	at scala.None$.get(Option.scala:529)
[error] 	at scala.None$.get(Option.scala:527)
[error] 	at $96fd4c6056e5b3f0e4bc$.$anonfun$microsite$20(build.sbt:244)
[error] 	at scala.Function1.$anonfun$compose$1(Function1.scala:49)
[error] 	at sbt.internal.util.$tilde$greater.$anonfun$$u2219$1(TypeFunctions.scala:62)
[error] 	at sbt.std.Transform$$anon$4.work(Transform.scala:68)
[error] 	at sbt.Execute.$anonfun$submit$2(Execute.scala:282)
[error] 	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:23)
[error] 	at sbt.Execute.work(Execute.scala:291)
[error] 	at sbt.Execute.$anonfun$submit$1(Execute.scala:282)
[error] 	at sbt.ConcurrentRestrictions$$anon$4.$anonfun$submitValid$1(ConcurrentRestrictions.scala:265)
[error] 	at sbt.CompletionService$$anon$2.call(CompletionService.scala:64)
[error] 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[error] 	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
[error] 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[error] 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[error] 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
[error] 	at java.lang.Thread.run(Thread.java:748)
[error] (Scalaunidoc / unidoc / scalacOptions) java.util.NoSuchElementException: None.get
```

I tried to downgrade unidoc plugin but nothing hepls. Google knows little about it 🤷

.github/workflows/test.yml
```diff
@@ -0,0 +1,49 @@
...
+
+jobs:
+  build:
+    name: Build and Test
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        java:
+          - adopt@1.8
+          - graalvm-ce-java11@20.3.0
```
BalmungSan, line 17: I believe it would be good to ensure the project works on multiple jdk versions. This one is using the JDK 11 but with **Graal** so it is like two birds with one stone.
What do you think?

dimafeng: yeah, totally agree!

.github/workflows/test.yml
```diff
@@ -0,0 +1,72 @@
...
+
+      - name: Build the microsite
+        run: ./sbt microsite/makeMicrosite
+
+      - name: Upload site data
+        uses: actions/upload-artifact@v2
+        with:
+          name: site
+          path: site/target/site
+          retention-days: 1
```
BalmungSan, line 56: Since this _" artefact"_ is only used for sharing data between the two jobs, I believe it makes sense to only retain it for one day instead of for `90` which is the default.

BalmungSan:
Actually I was thinking that we should not upload the site that was built but rather rebuilt it from source but not from the current commit but the last tag, would that be possible @dimafeng ?

dimafeng:
@BalmungSan looks like it's possible. Let me try

BalmungSan:
@dimafeng it seems the error is that the repository of the **sbt-launcher** changed to `https://repo.scala-sbt.org/scalasbt/maven-releases/org/scala-sbt/sbt-launch/`. So we may just go ahead a merge this as it is _(uncommenting the lines of course)_ and leave it fail during the next couple of days until we finish everything for the release of `0.16.0`.
What do you think?

dimafeng:
@BalmungSan I'm ok with it

BalmungSan (approved):

PR #202, opened 2020-11-14
Add Changelog

BalmungSan:
Our users deserved it!

----

Ok, I think I covered enough versions, previous releases are just too old for anyone to care IMHO.
Still, this has to be the last PR to merge before the `0.16.0`  release, to ensure all the dates, links and examples are OK.

PR #221, opened 2020-12-08
Removing Session & Using the new Rx module for Streaming

### LINKED ISSUE #149: Provide wrappers for the new Rx module

BalmungSan:
Fixes #149 Supersedes #164 Inspired by #67 

Ok, this is going to be a long PR. But hopefully, it will make things better for all.
The Scope of this PR will be removing **Session** for the library and instead create **Transactions** directly from the **Driver**. Also, I will replace our old _Streaming_ API with a new one based on the new `RxTransactions`.

- [x] Remove the old _Streaming_ API.
- [x] Remove **Session**.
- [x] Add a new _Straming_ API based on `RxTransaction`.
- [x] Update / fix the tests.
- [x] Implement the new **Stream** typeclass for all the _Streaming_ modules.
- [x] Update the docs.


bengraygh:
Great detective work, @BalmungSan! Thank you for investing a huge amount of time and effort to track down the root cause of this. This explains a lot of sporadic behaviour we have been seeing in our tests. To clarify, is the intention to use (but not expose) `Session` internally? My (possibly incorrect) understanding is that we still need it for the ability to chain causally linked transactions. Another question - does it make sense to fix the session issue in a separate PR for granularity? I can contribute with this issue.

BalmungSan:
Hi @bengraygh 

> This explains a lot of sporadic behaviour we have been seeing in our tests

Probably, hopefully, this will also fix that :)

> To clarify, is the intention to use (but not expose) Session internally

Yeah, we still use a **Session** internally since that is the only way to create **Transactions**, but at least for now, it won't be exposed in any form _(so no way to have casual chaining)_. Future work would be to see what would be the best way of exposing it.

I already pushed that change so you can take a look at how it looks like.

> Does it make sense to fix the session issue in a separate PR.

I thought about that, but since this change will break a lot of tests and the change of new _Streaming_ module will also break tests; and that last change also affect the final form of this change I decided that the most pragmatic solution, in this case, was doing both changes in one PR.
Now, you are totally correct that the _"best"_ ™️ in terms of traceability would have been doing two PRs.

BalmungSan:
OK, I am halfway through the tests but before continuing I realized that I made a design decision which I should have open up for debate. Thus, since it is still feasible to change it I will open a poll right now.

The problem is very simple, since **StreamingDriver** extends **Driver** should we override `transaction` so it returns a **StreamingTransaction** under the hood or should we leave it return a normal **Transaction**.

The first option implies that if you have a **StreamingDriver** then all operations will always be executed using the `Rx` API under the hood, so even operations like `single` would be creating a _Stream_ which will be reduced to a single element. The second one implies that even if you have a **StreamingDriver** most operations _(expect `stream` obviously)_ would be executed using the `Async` API; unless you call `streamingTransact` explicitly.

Vote with:
+ 👍  for the first option.
+ 🚀 for the second option.

core/src/main/scala/neotypes/Driver.scala
```diff
@@ -88,9 +88,46 @@ object Driver {
...
+              if (produced) {
+                s.onComplete()
+              } else {
+                s.onError(exceptions.TransactionWasNotCreatedException)
+              }
+            }
+          })
+        }
+      }
+      S.fromRx(unitPublisher).evalMap(_ => streamingTx)
```
BalmungSan, line 129: I hate this, period.
But at least works... hope https://github.com/neo4j/neo4j-java-driver/issues/797 gets fixed soon.

core/src/main/scala/neotypes/Driver.scala
```diff
@@ -2,89 +2,134 @@ package neotypes
...
+          val f = rxs.getClass.getDeclaredField("session")
+          f.setAccessible(true)
+          val ns = f.get(rxs).asInstanceOf[org.neo4j.driver.internal.async.NetworkSession]
+          ns.beginTransactionAsync(transactionConfig).accept(cb) { tx =>
+            val rxtx = new org.neo4j.driver.internal.reactive.InternalRxTransaction(tx)
+            Right(Transaction(rxtx, s))
+          }
+        }
+      }
+      S.fromF(streamingTx)
```
BalmungSan, line 115: I hate this, period.
But at least works... hope https://github.com/neo4j/neo4j-java-driver/issues/797 gets fixed soon.

dimafeng:
@BalmungSan I'll take a look over the weekend

.github/workflows/test.yml
```diff
@@ -36,7 +36,10 @@ jobs:
           restore-keys: |
             cache-v1-
 
-      - name: Compile and Test
+      - name: Compile
```
dimafeng, line 39: I'm just curious - is this done to make it fail faster if code doesn't compile in one of the scala versions?

build.sbt
```diff
@@ -204,6 +213,7 @@ lazy val monixStream = (project in file("monix-stream"))
       "org.typelevel" %% "cats-core" % catsVersion,
       "org.typelevel" %% "cats-effect" % catsEffectsVersion,
       "io.monix" %% "monix-eval" % monixVersion,
+      "io.monix" %% "monix-reactive" % monixVersion,
```
dimafeng, line 216: Does this duplicate line 217 or it's a github glitch?

core/src/main/scala/neotypes/Driver.scala
```diff
@@ -2,89 +2,134 @@ package neotypes
...
+        s.beginTransaction(transactionConfig).toStream[S].mapS { tx =>
+          Transaction[S, F](tx, s)
+        }
+      }
+      */
+
+      // Workaround for neo4j-java-driver#797 -------------------------------------------
+      val streamingTx = session.flatMap { s =>
+        F.async[StreamingTransaction[S, F]] { cb =>
+          val rxs = s.asInstanceOf[org.neo4j.driver.internal.reactive.InternalRxSession]
```
dimafeng, line 105: I know it's a temporary solution, but would it makes to move this reflection magic to the driver initialization phase (constructor)?

site/src/main/mdoc/overview.md
```diff
@@ -16,117 +16,116 @@ position: 10
 ## Session creation
 
 **neotypes** provides the `GraphDatabase` **factory** for creating a **Neo4j** `Driver`, which represents a connection with the Database.
-You can use this `Driver`, to create a `Session`, which can be used to perform operations _(`Transactions`)_ over the Database.<br>
-This **Scala** classes, are nothing more than simple wrappers over their **Java** counterparts. Which provide a more _"Scala-friendly"_ and functional API.
+You can use this `Driver` to perform operations _(`Transactions`)_ over the Database.<br>
+Those **Scala** classes, are nothing more than simple wrappers over their **Java** counterparts. But providing a more _"Scala-friendly"_ and functional API.
```
dimafeng, line 20: Minor grammar things (but I'm not 100% confident 😁 )
1. the comma in `classes, are` should be removed. 
2. `. But providing ...` is part of the previous sentence

site/src/main/mdoc/overview.md
```diff
@@ -16,117 +16,116 @@ position: 10
...
+val driver = GraphDatabase.driver[Future]("bolt://localhost:7687", AuthTokens.basic("neo4j", "****"))
+
+val result: Future[List[Movie]] =
+  """MATCH (movie: Movie)
+     WHERE lower(movie.title) CONTAINS "thing"
+     RETURN movie""".query[Movie].list(driver)
 ```
 
+Please remember that, you have to make sure that the `Driver` is properly closed at the end of the application execution, to make sure all obtained resources _(such as network connections)_ are cleaned up properly.<br>
+**Note**: for other effect types instead of `Future`, for example `IO`. Then, the creation of the `Driver` will be wrapped over some kind of **Resource**.
```
dimafeng, line 44: An optional, minor suggestion here
"for effect types other than `Future` (e.g. `IO`), the creation of the `Driver` will be managed by the effect specific implementation of **Resource**"

dimafeng:
That is a huge work! I couldn't go too deep but overall looks great to me! I left several minor comments but I'm ok to merge as-is.

.github/workflows/test.yml
BalmungSan: At first, it was an attempt to avoid **Travis** to kill the job because a single step took more than 40 minutes.

Then, I kind of liked the idea of making sure it compiled before running tests. Because I sometimes forget to run `+compile` or `+test` in local and finding that out faster seemed like a good idea.

But I am not too attached to it.

build.sbt
BalmungSan: I hope lol.

Will check tomorrow once I am on a computes again.

core/src/main/scala/neotypes/Driver.scala
BalmungSan: I prefer to keep all that is related to this _"solution"_ in a single place with the correct code commented.

That way it would be easier to change it once this is fixed.

build.sbt
BalmungSan: Indeed it was duplicated, sorry about that!
Maybe it happened during the `merge main`?

Anyways, it strikes me that **sbt** doesn't identify that as an error, but at least it is fixed now!
Thanks!

.github/workflows/test.yml
dimafeng: No, it makes sense. Let's leave it this way.

dimafeng:
LGTM! Let me know if you'd like me to merge and release

BalmungSan:
@dimafeng if it looks good to you then yes please!

I will finish the release notes in #202 after this is merged so you can take it a look during the week / weekend and release `0.16.0` whenever you can!

PR #167, opened 2020-08-12
WIP Unify Tests To WordSpec

geoffjohn11:

PR #259, opened 2021-02-14
initial commit

geoffjohn11:


BalmungSan:
Does this supersede #167?

core/src/test/scala/neotypes/CleaningIntegrationSpec.scala
```diff
@@ -2,7 +2,6 @@ package neotypes
 
 import org.scalatest.FutureOutcome
 
-/** Base class for integration specs that require to clean the graph after each test. */
```
BalmungSan, line 5: Why did you remove this comment?

core/src/test/scala/neotypes/StreamSuite.scala
```diff
@@ -51,6 +51,25 @@ abstract class StreamingDriverProvider[S[_], F[_]](testkit: StreamTestkit[S, F])
...
+//
+//  protected final def executeAsFutureList[A](work: DriverType => S[A]): Future[List[A]] =
+//    executeAsFuture(work andThen streamToFList)
+//
+//  protected final def debugMetrics(): F[Unit] =
+//    F.map(driver.metrics) { metrics =>
+//      println(s"METRICS: ${metrics}")
+//    }
+//}
+
```
BalmungSan, line 72: If this code is not needed anymore, then the best is to remove it, not just comment it.

BalmungSan (commented):
In general, I like it and thank you very much for all this work!

A couple of things:

1. About style check what we discussed in #167 I found some extra and missing whitespaces, most tests are not separated by an empty line _(that helps reading the code)_, and other minor details. However, do not worry too much about this since I probably will pull this branch and do a check and fix file by file just before merging. But, the more you do after that, the easier it will be for me :)
2. I found a couple of `should ===` which I would prefer if we always use `shouldBe`.
3. I think some tests can take advantage of other features of **ScalaTest**. Like `Inside` for when we access multiple parts of a case class or a tuple, or some of the collection matches like `theSameElementsAs`
4. I think it is better to extend the `Matchers` instead of importing them.
5. **Scalatest** recommends defining an abstract class _(or a trait)_ that extends all the things that will be common across all tests, and then extending this on all tests. In our case it would be the word spec and matches, that would help with the previous point.

Also, it may be good to check if we could use the typed triple equals _(and maybe plus artima supersafe)_ to ensure all our assertions will not always fail due to a type mismatch.
But that is a job for another separate PR.

core/src/test/scala/neotypes/AlgorithmSpec.scala
```diff
@@ -51,7 +49,7 @@ final class AlgorithmSpec[F[_]](testkit: EffectTestkit[F]) extends AsyncDriverPr
                 }
               )
            """.query[Unit].execute(d)
-      result <- """CALL gds.triangleCount.mutate('myGraph', {
+        result <- """CALL gds.triangleCount.mutate('myGraph', {
                      mutateProperty: 'tc'
```
geoffjohn11, line 53: @BalmungSan the formatting gets tough.  I'm not entirely sure how the extra whitespace is getting added.  Do you think the best way to make the formatting uniform across the project would be to add a scalafmt rule?  This would also ease the formatting burden as new contributors come on to the project and make these same mistakes.

BalmungSan: @geoffjohn11 would you like to try to add such formatting rules? I guess that would be great most popular projects do that.

geoffjohn11: I messed around this weekend with a few different format files.  None of them really came close to adhering to the format that already exists across the project.

BalmungSan: @geoffjohn11 yeah I know, but I guess it doesn't matter as long as it is a standard format.
I guess the best would be to open a new PR adding that so we can review how the codebase would change and after we merge that one update this one.

PR #281, opened 2021-03-08
Adding version scheme

BalmungSan:
Based on https://scala-lang.org/blog/2021/02/16/preventing-version-conflicts-with-versionscheme.html#versionscheme-librarydependencyschemes-and-sbt-150
More info here: https://www.scala-sbt.org/1.x/docs/Publishing.html#Version+scheme

It may even be good to release a new version after merging this, so it is included in the **pom**; version `0.16.1`
what do you think @dimafeng ?


dimafeng (approved):
Sure, I can merge and release 0.16.1

PR #204, opened 2020-11-14
WIP: Create query (orm-ish approach)

iRevive:
This PR brings a new module `neotypes-extra`.

These changes built on top of #201. Therefore until we merge #201 this PR duplicates some changes.

PR #294, opened 2021-03-29
Adding UnwrappedMappers semiauto derivation

### LINKED ISSUE #288: Automatic wrapping/unwrapping of `AnyVal`
Many libraries, such as circe, json4s or caliban, provide automatic wrapping and unwrapping of the inner value in `AnyVal` types without explicit mapping. Eg. `case class Id(value: String) extends AnyVal` will treated as a `String` for serialization and a string value would automatically be wrapped in an `Id` where required during deserialization.

It would be great if this were supported by neotypes as well.

BalmungSan:
Fixes #288

This probably should be moved into the `extras` module proposed in #204


dimafeng (approved):
Looks great! I can merge and release tomorrow in the morning.

BalmungSan:
@dimafeng yes please do!

PR #325, opened 2021-04-29
Add cypher interpolation of DeferredQueryBuilder

tjarvstrand:


core/src/main/scala/neotypes/CypherStringInterpolator.scala
```diff
@@ -17,25 +18,33 @@ object CypherStringInterpolator {
     val params = parameters.iterator
 
     @annotation.tailrec
-    def loop(paramNext: Boolean, output: List[DeferredQueryBuilder.Part]): List[DeferredQueryBuilder.Part] =
+    def loop(paramNext: Boolean, nextSubQuery: Int, output: List[DeferredQueryBuilder.Part]): List[DeferredQueryBuilder.Part] =
```
BalmungSan, line 21: Not related to your change, but since this method does a lot of concatenations I have thought of changing it to internally use either a `Vector` or a `ListBuilder`; of course, the return will still be a List`
WDYT?

core/src/main/scala/neotypes/CypherStringInterpolator.scala
```diff
@@ -17,25 +18,33 @@ object CypherStringInterpolator {
...
       if (paramNext && params.hasNext) {
         params.next() match {
-          case QueryArg.Param(value)      => loop(paramNext = false, output :+ DeferredQueryBuilder.Param(value))
-          case QueryArg.CaseClass(params) => loop(paramNext = false, output ++ caseClassParts(params))
+          case QueryArg.Param(value)          => loop(paramNext = false, nextSubQuery, output :+ DeferredQueryBuilder.Param(value))
+          case QueryArg.CaseClass(params)     => loop(paramNext = false, nextSubQuery, output ++ caseClassParts(params))
+          case QueryArg.QueryBuilder(builder) => loop(paramNext = false, nextSubQuery + 1, output ++ subQueryParts(builder, nextSubQuery))
```
BalmungSan, line 26: I think it would be good to use a named parameter for `index = nextSubQuery`

core/src/main/scala/neotypes/CypherStringInterpolator.scala
```diff
@@ -17,25 +18,33 @@ object CypherStringInterpolator {
...
         }
       } else if (queries.hasNext) {
-        loop(paramNext = true, output :+ DeferredQueryBuilder.Query(queries.next()))
+        loop(paramNext = true, nextSubQuery, output :+ DeferredQueryBuilder.Query(queries.next()))
       } else {
         output
       }
 
-    val queryParts = loop(paramNext = false, Nil)
+    val queryParts = loop(paramNext = false, 1, Nil)
```
BalmungSan, line 34: It would be good to use named parameters for this initial call to `loop` and maybe rename `output` to `acc`

core/src/main/scala/neotypes/CypherStringInterpolator.scala
```diff
@@ -81,14 +90,15 @@ sealed trait QueryArg
 object QueryArg {
   final case class Param(value: QueryParam) extends QueryArg
   final case class CaseClass(params: Map[String, QueryParam]) extends QueryArg
+  final case class QueryBuilder[T](value: DeferredQueryBuilder) extends QueryArg
```
BalmungSan, line 93: Why the `T` type parameter?

core/src/main/scala/neotypes/CypherStringInterpolator.scala
```diff
@@ -81,14 +90,15 @@ sealed trait QueryArg
...
 }
 
 @annotation.implicitNotFound("Could not find the QueryArgMapper for ${A}.")
 trait QueryArgMapper[A] {
   def toArg(value: A): QueryArg
 }
 
-object QueryArgMapper extends QueryOrgMappersLowPriority {
+object QueryArgMapper extends QueryArgMappersLowPriority {
```
BalmungSan, line 101: :+1:

core/src/main/scala/neotypes/CypherStringInterpolator.scala
```diff
@@ -97,9 +107,11 @@ object QueryArgMapper extends QueryOrgMappersLowPriority {
 
 }
 
-trait QueryOrgMappersLowPriority {
+trait QueryArgMappersLowPriority {
   implicit def fromCaseClassArgMapper[A: CaseClassArgMapper]: QueryArgMapper[A] =
     CaseClassArgMapper[A]
+
+  implicit val deferredQueryBuilderArgMapper: QueryArgMapper[DeferredQueryBuilder] = QueryArg.QueryBuilder(_)
```
BalmungSan, line 114: Just for style consitency, I would move the body to the next line.

core/src/main/scala/neotypes/DeferredQuery.scala
```diff
@@ -358,14 +358,16 @@ private[neotypes] object DeferredQuery {
...
 
   /** Creates a [DeferredQuery] from this builder.
     *
     * @tparam T the result type of the constructed query.
     * @return A [DeferredQuery[T]]
     */
-  def query[T]: DeferredQuery[T] = {
+  def query[T]: DeferredQuery[T] = query(PARAMETER_NAME_PREFIX)
+
+  private[neotypes] def query[T](parameterNamePrefix: String): DeferredQuery[T] = {
```
BalmungSan, line 370: I do not understand the purpose of this change?

BalmungSan (approved):
Thanks! Just some minor notes.

core/src/main/scala/neotypes/CypherStringInterpolator.scala
tjarvstrand: Not sure. I must say that I sadly have very little concept of the performance implications. Coming from a world where mutability is not even an option I usually don't even consider `ListBuilder` :D Unless I'm wrong however this would only affect compile-time performance, right?

core/src/main/scala/neotypes/CypherStringInterpolator.scala
tjarvstrand: :+1:

core/src/main/scala/neotypes/CypherStringInterpolator.scala
tjarvstrand: Ah, leftover. Thanks

core/src/main/scala/neotypes/DeferredQuery.scala
tjarvstrand: The parameterized version is used by `CypherStringInterpolator.subQueryParts` to prefix subquery parameters so they don't collide...

But thinking about it made me realize that this will probably not work if there is more than one level of nested subqueries :/

core/src/main/scala/neotypes/CypherStringInterpolator.scala
BalmungSan: _(for some reason I can't tag you with @ oh well, hope you see it)_

> I usually don't even consider ListBuilder :D

Yeah, I also try to avoid mutability in my code, but a local-scoped `ListBuilder` doesn't hurt :)

> Unless I'm wrong however this would only affect compile-time performance, right?

But, I believe you are right. This is why I never bothered to make the change myself, I just decided to mention it today to see what you think.

core/src/main/scala/neotypes/DeferredQuery.scala
BalmungSan: That is a good test actually!

PR #341, opened 2021-05-17
Make default transaction config configurable

tjarvstrand:


core/src/main/scala/neotypes/Driver.scala
```diff
@@ -18,35 +18,44 @@ import scala.jdk.CollectionConverters._
...
   def streamingTransact[T](config: TransactionConfig)(txF: StreamingTransaction[S, F] => S[T]): S[T]
 
   final def streamingTransact[T](txF: StreamingTransaction[S, F] => S[T]): S[T] =
-    streamingTransact(config = TransactionConfig.default)(txF)
+    streamingTransact(config = transactionConfig)(txF)
+
+  /** Creates a new driver instance using the provided transaction configuration as default. NB. This does NOT create
+   * a new instance of the underlying Java driver.
+   */
+  def withTransactionConfig(config: TransactionConfig): StreamingDriver[S, F]
```
BalmungSan, line 58: I believe we could _(should?)_ move this **svaladoc** comment to the method defined in `Driver` and inherit the documentation here.
Also, IIRC, we can mark this method as `override` and still leave it abstract.

core/src/main/scala/neotypes/Driver.scala
```diff
@@ -55,7 +64,8 @@ object Driver {
     case (tx, Some(_)) => tx.rollback
   }
 
-  private class DriverImpl[F[_]](driver: NeoDriver)
+  private class DriverImpl[F[_]](driver: NeoDriver,
+                                 val transactionConfig: TransactionConfig = TransactionConfig.default)
```
BalmungSan, line 68: Please add `override`

core/src/main/scala/neotypes/Driver.scala
```diff
@@ -79,11 +89,18 @@ object Driver {
       F.async { cb =>
         driver.closeAsync().acceptVoid(cb)
       }
+
+    /** Creates a new driver instance using the provided transaction configuration as default. NB. This does NOT create
+     * a new instance of the underlying Java driver.
+     */
+    def withTransactionConfig(config: TransactionConfig): Driver[F] = new DriverImpl(driver, config)
```
BalmungSan, line 96: This class is private so we do not need to add a **scaladoc** comment here.

core/src/main/scala/neotypes/Driver.scala
```diff
@@ -123,6 +140,12 @@ object Driver {
 
       S.resource(tx)(txF)(txFinalizer)
     }
+
+    /** Creates a new driver instance using the provided transaction configuration as default. NB. This does NOT create
+     * a new instance of the underlying Java driver.
+     */
+    override def withTransactionConfig(config: TransactionConfig): StreamingDriver[S, F] =
+      new StreamingDriverImpl(driver, config)
```
BalmungSan, line 148: ditto.

BalmungSan (approved):
Thanks!

Just some minor notes.

BalmungSan (approved):

PR #344, opened 2021-05-17
Add cypher interpolation of DeferredQueryBuilder

tjarvstrand:


core/src/main/scala/neotypes/DeferredQuery.scala
```diff
@@ -367,44 +378,64 @@ final class DeferredQueryBuilder private[neotypes] (private val parts: List[Defe
     */
   def query[T]: DeferredQuery[T] = {
     @annotation.tailrec
-    def loop(remaining: List[Part], queryBuilder: StringBuilder, accParams: Map[String, QueryParam], nextParamIdx: Int): DeferredQuery[T] =
+    def loop(remaining: List[Part],
+             queryBuilder: StringBuilder,
```
BalmungSan, line 382: I think we should move the `queryBuilder` argument outside of the `loop`, it is mutable in any case.

core/src/main/scala/neotypes/DeferredQuery.scala
```diff
@@ -323,6 +325,15 @@ final case class DeferredQuery[T] private[neotypes] (query: String, params: Map[
     */
   def withParams(params: Map[String, QueryParam]): DeferredQuery[T] =
     this.copy(params = this.params ++ params)
+
+  private[neotypes] def withParameterPrefix(prefix: String): DeferredQuery[T] = {
```
BalmungSan, line 329: Having this method in this class feels wrong. But I understand that otherwise, we would need to duplicate logic or move this work that right now is done at compile time _(which is great)_ to runtime _(which would be worse to users)_.

What about if we move both the `query[T]` method as well as this one to a helper object?
Especially because once the final query is built, these param locations are actually unnecessary, right?

core/src/main/scala/neotypes/CypherStringInterpolator.scala
```diff
@@ -17,19 +17,20 @@ object CypherStringInterpolator {
...
       } else if (queries.hasNext) {
-        loop(paramNext = true, output :+ DeferredQueryBuilder.Query(queries.next()))
+        loop(paramNext = true, nextSubQuery, acc :+ DeferredQueryBuilder.Query(queries.next(), Seq.empty))
       } else {
-        output
+        acc
       }
 
-    val queryParts = loop(paramNext = false, Nil)
+    val queryParts = loop(paramNext = false, nextSubQuery = 1, Nil)
```
BalmungSan, line 33: Could you please add `acc =` before the `Nil`?
Also, as we discussed the other day, maybe it would be better to use `Vector` due to the constant appending?
While I understand this is run at compile-time and queries are in general _"small"_ so complexity shouldn't matter that much, I wonder if we are introducing unnecessary slowdowns to codebases that have a lot of queries.

BalmungSan (commented):
Hi, @tjarvstrand thanks for giving another go to this one.
I like the capabilities this would bring to the library, I have some minor concerns with the implementation; but nothing strong enough to avoid merging this.

Also, since this is a more involved change I would also like to have a review from @dimafeng

----


BTW, maybe it would be good to have a test with two subqueries and a test with a subquery that, in turn, also has a subquery.
WDYT?

Finally, it would be good to add a note about this to the docs; would you be able to also do that?
I can do it after the merge if you prefer.

core/src/main/scala/neotypes/DeferredQuery.scala
tjarvstrand: I actually disagree with that. It deals with the internal structure of the query so is the correct place for it if you ask me. If you want me to move it I can do that though. Just tell me where :)

core/src/main/scala/neotypes/CypherStringInterpolator.scala
tjarvstrand: > Could you please add acc = before the Nil?

Sure

>While I understand this is run at compile-time and queries are in general "small" so complexity shouldn't matter that much, I wonder if we are introducing unnecessary slowdowns to codebases that have a lot of queries.

Perhaps, but IMHO there is no point in optimizing for problems that you don't know exist. I'm also not sure that is a change which would belong in this PR

core/src/main/scala/neotypes/DeferredQuery.scala
tjarvstrand: Sure, I can do that. It's a bit weird right now. Feels like a leftover from something else

core/src/main/scala/neotypes/CypherStringInterpolator.scala
tjarvstrand: >Perhaps, but IMHO optimizing for problems that you don't know exist is usually an unnecessary time sink. I'm also not sure that is a change which would belong in this PR

I'm a bit tired atm :P I realized it's a really small so I can of course do it if you want. I'm not really sure what's the better option though; using a `mutable.List` or just prepending the parts and reversing the list at the end

tjarvstrand:
> BTW, maybe it would be good to have a test with two subqueries and a test with a subquery that, in turn, also has a subquery.
WDYT?

Unless I misunderstand, there is such a test in `CypherStringInterpolatorSpec` :)

> Finally, it would be good to add a note about this to the docs; would you be able to also do that?

Sure, I can do that

core/src/main/scala/neotypes/DeferredQuery.scala
BalmungSan: > No, they need to be kept so that parameter names can be prefixed if the query later gets nested in another one

Yeah, that is what I tried to mean _"final query"_.

The reason I do not like it here is that we do not want to embed a `DefferedQuery` inside a `DefferedQueryBuilder`, we want to embed a `DefferedQueryBuilder` inside another one; that is why having to call `query` for the builder in order to embed it in another builder feels weird for me. 
Thus, I think it would be good to move both this method as well as the loop inside `query` to a helper object. But do not ask me for naming, I suck at that 😅

Does that make more sense to you? Or what do you think?

core/src/main/scala/neotypes/CypherStringInterpolator.scala
BalmungSan: > using a mutable.List or just prepending the parts instead of appending and then reversing the list at the end

Yeah, there is also another reason why every time I think about this I never actually do anything; there are many ways to tackle this. I just wanted a second opinion about this.
Lets agree that this change doesn't below to this PR and leave the idea in the air in case you want to chat about in in **gitter**.

BalmungSan:
> Unless I misunderstand, there is such a test in CypherStringInterpolatorSpec :)

I see a test for a subquery that has two parameters.
I was thinking more in these two cases:

```scala
// Two subqueries.
val subQuery1Param = 1
val subQuery1 = c"user.id = ${subQuery1Param}"
val subQuery2Param = "Luis"
val subQuery2 = c"user.name = ${subQuery2Param}"
val query = c"MATCH (user: User) WHERE ${subQuery1} OR ${subQuery2} RETURN user"

// Subquery with a subquery.
val subSubQueryParam = 1
val subSubQuery = c"user.id = ${subSubQueryParam}"
val subQuery = c"""${subSubQuery} OR user.name = "Luis""""
val query = c"MATCH (user: User) WHERE ${subQuery} RETURN user"
```

Do you think such tests would be helpful?

> Sure, I can do that

🚀

core/src/main/scala/neotypes/DeferredQuery.scala
tjarvstrand: > The reason I do not like it here is that we do not want to embed a DefferedQuery inside a DefferedQueryBuilder, we want to embed a DefferedQueryBuilder inside another one...

I'm not sure that's possible. At least, I wouldn't know how to do it.

core/src/main/scala/neotypes/CypherStringInterpolator.scala
tjarvstrand: Fine by me :)

tjarvstrand:
>Do you think such tests would be helpful?

Gotcha. I added a more involved testcase :)

core/src/main/scala/neotypes/DeferredQuery.scala
BalmungSan: Uhm, let's just leave it as it is right and I will try to play with the code later.
In any case, I am no even sure if my idea makes sense since I haven't tried to code it :)

BalmungSan (approved):

BalmungSan:
Thanks for the hard work @tjarvstrand I see it ok as it is right now :+1:
I will wait a couple of days in case @dimafeng would be able to also give it a review before merging it!

dimafeng:
@tjarvstrand thanks for your contribution! 🚀

PR #345, opened 2021-05-17
Ensuring all casting errors during mapping are encapsulated in a NeotypesException

### LINKED ISSUE #343: Wrap UUID parsing exceptions into Neotypes exceptions
UUIDs are mapped using this ValueMapper:

https://github.com/neotypes/neotypes/blob/3a2edd01ff5ff13cfb2c0c9c61a4745f72b43738/core/src/main/scala/neotypes/mappers.scala#L535-L536

But `UUID.fromString` throws `IllegalArgumentException`s, `NumberFormatException`s and probably other types of exceptions which are not instances of `NeotypesException`. Ultimately, these exceptions will be returned by the library and I think it would be nicer if they were included in the `NeotypesException` hierarchy in order to allow the user to implement reliable logic when these errors appear.

BalmungSan:
Fixes #343


dimafeng (approved):
🚀

PR #362, opened 2021-06-16
Don't override driver's tx-config from DeferredQuery

tjarvstrand:


BalmungSan (approved):
Oh right, this doesn't play well with #341

I think we could release a new version just with this fix and to ensure the appropriate changelog is published.

tjarvstrand:
Thanks for the quick turnaround!

PR #364, opened 2021-06-17
Reverting the neo4j#797 workaround

BalmungSan:
https://github.com/neo4j/neo4j-java-driver/pull/932 fixed https://github.com/neo4j/neo4j-java-driver/issues/797

Thanks to that we can revert the ugly workaround we had and rather use the pretty code we all liked.

Of course, we need to wait for a proper `4.3.2` release of the **neo4j-java-driver** before we can merge this one.
However, I wanted to open this PR right now to discuss something.

Since we do not enforce the version of the driver since it is provided and then it is the responsibility of our users to upgrade the driver to the latest patch version for this code to work. If they update to a newer version of **neotypes** without upgrading the driver then they will have runtime errors; which is not nice.

I believe that we must, at least, be extremely explicit about it somewhere in the docs.
But maybe we may also keep the previous workaround just for compatibility? Since we are not still at `1.0.0` I think its fine to just require the latest versions of the underlying driver; especially since it seems they will backport it to previous versions like `4.2` and `4.1` so users would only need to update a patch version, not a minor version which may be more problematic.

What do other folks think about this?
@dimafeng @tjarvstrand @iRevive @geoffjohn11

### DIFF
--- a/core/src/main/scala/neotypes/Driver.scala
+++ b/core/src/main/scala/neotypes/Driver.scala
@@ -106,29 +106,11 @@ object Driver {
         driver.rxSession(sessionConfig)
       }
 
-      /* Use when https://github.com/neo4j/neo4j-java-driver/issues/797 is fixed.
       S.fromF(session).flatMapS { s =>
         s.beginTransaction(transactionConfig).toStream[S].mapS { tx =>
           Transaction[S, F](tx, s)
         }
       }
-      */
-
-      // Workaround for neo4j-java-driver#797 -------------------------------------------
-      val streamingTx = session.flatMap { s =>
-        F.async[StreamingTransaction[S, F]] { cb =>
-          val rxs = s.asInstanceOf[org.neo4j.driver.internal.reactive.InternalRxSession]
-          val f = rxs.getClass.getDeclaredField("session")
-          f.setAccessible(true)
-          val ns = f.get(rxs).asInstanceOf[org.neo4j.driver.internal.async.NetworkSession]
-          ns.beginTransactionAsync(transactionConfig).accept(cb) { tx =>
-            val rxtx = new org.neo4j.driver.internal.reactive.InternalRxTransaction(tx)
-            Right(Transaction(rxtx, s))
-          }
-        }
-      }
-      S.fromF(streamingTx)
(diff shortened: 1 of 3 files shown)


tjarvstrand:
> Since we do not enforce the version of the driver since it is provided

Even if it is provided you can still require a minimum version right? I'm a firm believer in minimizing room for errors so if forcing the minimum version is an option I think that is definitely the way to go.

michael-simons:
FWIW from a Spring persons perspective: We do a lot of dependency management. The version we define in dependency management is the minimum version required to function. We guarantee that later also do work (to some extend, not including major versions, but patch and minors).

BalmungSan:
> Even if it is provided you can still require a minimum version right?
> The version we define in dependency management is the minimum version required to function

@tjarvstrand @michael-simons are you two referring as a convention? Because I also agree with requiring the latest patches to work is OK.
The problem is that, AFAIK, there is no way to enforce that from tooling, users can always use the version of the underlying **Java** driver they want. Now, one can say that if you used an unsupported version and you got errors is your fault; that is why I think the best solution is to just be very explicit about the minimum version in the docs.

I just wanted to confirm that this would not be a problem for current users.
In any case, from what I have seen, newer versions of the **Java** driver basically provide support for newer versions of **neo4j** and bump dependencies, so I do not think it is a big deal to enforce users to upgrade even if it would have been to a minor release.

michael-simons:
At least in the Java world: It's hard to enforce. People can always downgrade, but then they are on an unsupported path.

> In any case, from what I have seen, newer versions of the Java driver basically provide support for newer versions of neo4j and bump dependencies

Yes, that's the case for 4.1, 4.2 and 4.3. 
5.0 will very likely see breaking changes, which we will try to announce with deprecations ins 4.4

iRevive:
I'm fine with this change. 

Perhaps we can have some kind of a runtime warning, something that Akka has for a while.

BalmungSan:
@tjarvstrand @michael-simons @iRevive thanks for the feedback!

I will proceed then with releasing `0.18.3` which will only include this fix, the changelog and the main page of the docs include notes about the implication of this release.
@dimafeng can you help us with the release?

PR #375, opened 2021-07-07
Using coursier/cache-action

BalmungSan:
@dimafeng I think is better to use this since it takes care of all details for us, like it also considers the files inside `project/` which we did not.


dimafeng:
@BalmungSan nice! 💯

PR #388, opened 2021-08-02
Using the official sbt script

BalmungSan:
This PR replaces the **sbt-extras** launcher script with the official one.
The motivation is that I would rather have the official one; especially since AFAIK we were not using any of the features of the **sbt-extras** one, and that the only reason for using it was that the official one was not easily _curleable_ before.

Even more, I wonder what is even the need for that script @dimafeng? I kind of remember a discussion in **gitter** where you said the idea was for newcomers to don't need to install **sbt**, but I wonder if that is really a good motivation nowadays when we have things like **sdkman** and **coursier** that make the installation really simple, or that downloading that script yourself is really simple: `curl https://raw.githubusercontent.com/sbt/sbt/develop/sbt >> sbt`
So maybe, we should rather just remove it? In any case, the **setup-scala** Github action installs the script for use during CI.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Setup Java
-        uses: olafurpg/setup-scala@v10
+        uses: olafurpg/setup-scala@v13
         with:
           java-version: ${{ matrix.java }}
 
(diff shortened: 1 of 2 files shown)


dimafeng:
@BalmungSan yeah, I'm fine with removing it but I'm gonna merge this one meanwhile 😁

PR #405, opened 2021-08-31
Streaming Guarantee Spec initial commit

geoffjohn11:


akka-stream/src/main/scala/neotypes/akkastreams/AkkaStreams.scala
```diff
@@ -34,15 +33,18 @@ trait AkkaStreams {
       override def fromF[A](future: Future[A]): AkkaStream[A] =
         Source.future(future)
 
-      override final def resource[A, B](r: Future[A])(f: A => AkkaStream[B])(finalizer: (A, Option[Throwable]) => Future[Unit]): AkkaStream[B] =
+      override final def guarantee[A, B](r: Future[A])
+                                        (f: A => AkkaStream[B])
+                                        (finalizer: (A, Option[Throwable]) => Future[Unit]): AkkaStream[B] =
         Source.future(r).flatMapConcat { a =>
-          f(a).watchTermination() {
```
geoffjohn11, line 39: @BalmungSan I made this change because this code doesn't handle a test case.  If the finalizer fails, the stream doesn't fail with the finalizer exception, which is how the other streaming implementations behave.  I'm not certain if this is exactly right, but it seemed like it might be an idea, and it passes the test case.

PR #440, opened 2021-11-19
Akka stream fix

geoffjohn11:


BalmungSan (approved):
@geoffjohn11 I really don't understand why it works but oh well, thanks!

PR #455, opened 2021-12-23
Rollback Transaction CancelSpec

### LINKED ISSUE #189: Properly test: should rollback on cancellation
During developing #164 I noticed that I made a big mistake on some of my previous implementations and it is that I am `committing` a **Transaction** when the effect was _cancelled_. I just realized that is an error, I believe most people would expect that an effect running some queries in the same **Transaction** to be `rollbacked` on cancellation.

However, #164 is too big and this change is also big enough so I decided to create this issue to not forget about this and fix it after finishing #164

geoffjohn11:
Fixes #189

### DIFF
--- a/zio/src/test/scala/neotypes/zio/ZioTaskSuite.scala
+++ b/zio/src/test/scala/neotypes/zio/ZioTaskSuite.scala
@@ -20,6 +20,9 @@ object ZioTaskTestkit extends EffectTestkit[Task] {
       override final def runConcurrently(a: Task[Unit], b: Task[Unit]): Task[Unit] =
         a.zipPar(b).map(_ => ())
 
+      override final def cancel[A](fa: Task[A]): Task[Unit] =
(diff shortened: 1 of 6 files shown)


zio/src/test/scala/neotypes/zio/ZioTaskSuite.scala
```diff
@@ -20,6 +20,9 @@ object ZioTaskTestkit extends EffectTestkit[Task] {
       override final def runConcurrently(a: Task[Unit], b: Task[Unit]): Task[Unit] =
         a.zipPar(b).map(_ => ())
 
+      override final def cancel[A](fa: Task[A]): Task[Unit] =
+        fa.fork.map(_.interrupt).flatten.unit
```
geoffjohn11, line 24: ZIO doesn't like flatMap, so changed to this

BalmungSan (approved):
Awsome!

PR #481, opened 2022-03-02
Update cats effect and fs2 to cats-effect 3 versions

Jacoby6000:
Note that Monix is still on CE2.  I don't  think that should be an issue though. I doubt people would depend on both monix and ce/fs2 modules of this project at the same time

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,8 +11,8 @@ val mockitoVersion = "1.10.19"
 val scalaTestVersion = "3.2.11"
 val logbackVersion = "1.2.10"
 val catsVersion = "2.7.0"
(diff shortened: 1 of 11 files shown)


BalmungSan:
@dimafeng I had to _"disable"_ all **monix** docs because of classpath problems; but I would guess we are fine now.

Feel free to merge this, fix the release date in the `CHANGELOG` and do a new release whenever you can and want :D

dimafeng:
@BalmungSan Awesome!!!

I just sent you a message in gitter -- I added you as an owner, feel free to merge and release at your convenience.

Issue #522, opened 2022-05-29
Request: ZIO 2.x support


PhilAndrew:
ZIO 2.x is going along https://zio.dev/next/howto/migrate/zio-2.x-migration-guide/

Request to support

BalmungSan:
Hi @PhilAndrew just like how we did for the **cats-effect** `3.x` release, we won't upgrade until the `2.x` is officially released _(since we can't provide both versions)_.
Also, a quick look at the migration guide suggests that we either could not upgrade or that our library won't be idiomatic for **ZIO** users _(if ever it was)_.

In any case, if you or anyone wants to open a PR to do the migration we are more than happy to review it and merge it when the time comes.

masonedmison:
@BalmungSan I am just curious about your take on what is different about v2.X that would make it any less idiomatic than v1.X? Is it the fact that `ZManaged` is no more? At any rate, I have been getting more familiar with ZIO for work and would be happy to try to add support for this once an official release is out for v2.  I _think_ the current release candidate version should be relatively stable so we might be able to get a headstart before it's officially released :)

BalmungSan:
@masonedmison

> I am just curious about your take on what is different about v2.X that would make it any less idiomatic than v1.X? Is it the fact that ZManaged is no more? 

Mostly yes, it seems we would need to look into `ZLayers` madness or see how to trick the type system into using the `R` type. But, again, I really haven't looked into that.

> At any rate, I have been getting more familiar with ZIO for work and would be happy to try to add support for this once an official release is out for v2

That would be awesome!

>  I think the current release candidate version should be relatively stable so we might be able to get a headstart before it's officially released :)

Sure a WIP PR is more than welcome!

masonedmison:
@BalmungSan Sounds good. Just started dinking around with it. As soon as I get something close-ish, I will open up a WIP PR :)

BalmungSan:
Thanks to the work of @masonedmison this is technically closed.
So now all that remains is when to merge and release?

I am not a **ZIO** user myself so I really don't care.
Thus, I am asking to all **ZIO** users of **neotypes** to vote, should we release now (🚀) or when the official `2.0` release is done (👀)?

PS: Note that I don't plan to support both versions so releasing now means no more **ZIO** `1.0` releases.

BalmungSan:
Closed by #524

PR #524, opened 2022-06-14
Zio2 support

masonedmison:
For request #522. This currently uses `2.0.0-RC6` as a placeholder for an official `2.0.0` release. For the most part, this was pretty mechanical, the only bits that might warrant a closer look are 1) `guarantee` and 2) `ZioResource`. Regarding 1, a `ZIO` is constructed with the acquire and release actions that requires a `Scope` in the environment - this scope is eliminated within the scope of this call (using `ZIO.scoped`). Regarding 2, we just return a `ZIO` that has a `Scope` part of the required environment. I _think_ both of these make sense given that I understand the desired semantics.


masonedmison:
Note that I haven't touched `zio-streams` yet - I assumed that this should be a separate PR but if you think it would make sense, I could take a stab at migrating the streams module as part of this PR, too.

BalmungSan:
> Note that I haven't touched zio-streams yet - I assumed that this should be a separate PR but if you think it would make sense, I could take a stab at migrating the streams module as part of this PR, too.

Ideally it should be a single PR IMHO, I don't like to break main :)

masonedmison:
> Ideally it should be a single PR IMHO, I don't like to break main :)
Fair and makes sense. I'll start working on `zio-streams` then :)

masonedmison:
Ok that was silly, it was pretty trivial to migrate `ZioStream` - for some reason I thought it was going to be more complicated

BalmungSan:
@masonedmison it seems the tests are green! 🎉 
What is failing now are the docs which are still mentioning things like `ZManaged` to generate the docs locally just run `sbt microsite/makeMicrosite`

BalmungSan (approved):
@masonedmison thanks a lot for looking into this.
I will keep the PR open until the official release; unless you think it would be better to do a release now?

BTW, while I will try to keep a look on when **ZIO** releases, but since you are real users of it I would appreciate if you can remember me once it happens.

masonedmison:
> @masonedmison thanks a lot for looking into this. I will keep the PR open until the official release; unless you think it would be better to do a release now?
@BalmungSan  My pleasure! I will try to contribute something more interesting and technically challenging next time. :) In terms of releasing now or later, I think it would be Ok to release now. I am not really clear on when to expect an official release but I'll circle back here though when the official v2 release is out and make any changes necessary.

PR #548, opened 2022-07-25
Separate subproject for generic

tjarvstrand:


BalmungSan:
So the only way to solve the cycles was to create new projects? Then I guess all that this PR is missing is marking those new modules as `publish := false`

BalmungSan (approved):
Thanks! :rocket:

PR #555, opened 2022-08-28
WIP Scala3 support for `generic` module and `CypherStringInterpolator`

masonedmison:
**Work in progress** So far I have first draft implementations of `QueryArgMapper` and `ResultMapper` derivation for both `semiauto` (including support for `derived <typeclass>`) and `auto` derivations.

Also made were some minor changes - things like removing `final` from `object`s as well as modifying `void` in `Stream` (the `S[_]` is not allowed in `void` in scala3 and the compiler complains with `unreducible application of higher-kinded type S to wildcard arguments`)

> For `auto`, the following import is needed `generic.auto.given` vs `generic.auto._` in the scala 2 implementation.

Todos for `generic`
- [ ] fix tests to not use `auto`. Auto-derivation in scala3 requires to import `given` like `import auto.given` so since this syntax will not work for both scala2 and scala3 I would maybe suggest we derive all case classes used in testing using semi-auto derivation so that tests can be run both using both versions.
- [ ] Ensure `Option` prioritization. works for `QueryArgMapper`.
- [x] Revisit implementation for `ResultMapper` to see if we can leverage `shapeless3` here more - currently the approach is very similar to that taken using `shapeless2` except using `Tuple` instead of `Generic`.

Todos for `CypherStringInterpolator`
- [x] seperate out non-macro stuff into shared `scala` directory
- [x] implement macro in `scala3`.

Todos - misc.
- [ ] Separate out shapeless specific tests to **scala2** test directory

I know this PR is still very rough, but I wanted to get some eyes on it sooner than later to get any feedback :).

Note that this is dependent on PR #584 as I am currently basing my work off of this branch.


build.sbt
```diff
@@ -61,9 +62,38 @@ def removeScalacOptionsInTest(scalaVersion: String) =
...
+  }
+
+def maybeReflect(scalaVersion: String) =
+  CrossVersion.partialVersion(scalaVersion) match {
+    case Some((2, _)) =>
+      Seq(
+        "org.scala-lang" % "scala-reflect" % scalaVersion
+      )
+    case _ => Seq.empty  
+  }
```
BalmungSan, line 84: Why not moving this to `macrosDepsByVersion`? Unless I am blinder than what I believe both are always sued together so I would rather just have one.
WDYT?

build.sbt
```diff
@@ -61,9 +62,38 @@ def removeScalacOptionsInTest(scalaVersion: String) =
...
+
+def compileOptsByVersion(scalaVersion: String) =
+  CrossVersion.partialVersion(scalaVersion) match {
+    case Some((2, _)) =>
+      Seq("-Ywarn-macros:after")
+    case _ => 
+      Seq.empty
+  }
```
BalmungSan, line 92: What about adding to this the `enablePartialUnificationIn2_12` logic I have below?

BalmungSan (commented):
I still haven't checked the deriving logic, but about the other changes, I have some questions / comments?

+ Why removing `final` from `objects` does it finally warn in **Scala 3**? IIRC in **Scala 2** it was needed to trigger some byte code optimizations because reasons.
+ Make sure to configure your IDE to always insert an EOL before the EOF.
+ What do you and other folks think of removing `2.12`? And just cross-compiling for `2.13` and `3`? That would simplify some bits, and we could remove the collection-compat library; but not sure if the tradeoff is worth it.

build.sbt
masonedmison: Sure, sounds good to me.

build.sbt
masonedmison: Sounds good, I will aggregate these "byVersion" functions.

masonedmison:
> Why removing final from objects does it finally warn in Scala 3? IIRC in Scala 2 it was needed to trigger some byte code optimizations because reasons.

In scala3 the compiler complains with an error of `Modifier final is redundant for this definition`. This _might_ be a nonFatal error that could be ignored with a compiler option if need be?

> Make sure to configure your IDE to always insert an EOL before the EOF.

Will do.

> What do you and other folks think of removing 2.12? And just cross-compiling for 2.13 and 3? That would simplify some bits, and we could remove the collection-compat library; but not sure if the tradeoff is worth it. 

Hrm, this is a tough one. If I am remembering right, didn't someone provide stats somewhat recently that showed that there are still a large number of monthly downloads for `2.12`? In general, I would have to think more about this - I am still very new to the world of OSS and would have to see how other libraries are handling `2.12`.

BalmungSan:
Hi @masonedmison I forgot if this one was already ready for a final review and merge, or if you were still working on fixing some tests or something.

masonedmison:
@BalmungSan Not ready for review yet. I am still working on porting the `CypherInterpolator` as well as fixing some tests. Things have gotten a little busy for me this past week but I hope to revisit this shortly. Apologies for the delay! :)

BalmungSan:
@masonedmison no worries, I was actually worried that this was blocked because of me.

build.sbt
masonedmison: I won't address this one as `enablePartialUnificationIn2_12` is only currently used in the `catsEffect` project to where as `compileOptsByVersion` is used as part of `commonSettings`.

masonedmison:
Just pushing everything up that I have worked on already and rebased onto `main`.

PR #584, opened 2022-12-11
WIP: Neotypes-Schema — Explicit decoders

### LINKED ISSUE #236: Discussion: Use explicit decoders instead of typeclass based ones
Following the ideas of other libraries like [**skunk**](https://tpolecat.github.io/skunk/index.html) and [**dynosaur**](https://systemfw.org/dynosaur/#/), I believe we should also provide explicit decoders instead of implicit ones.

Obviously, the downside of that is extra boilerplate for simple cases.
The idea would be to provide an API that allows to easily query and decode simple types like primitives and provide generic support for explicit derivation of decoders for **case classes** _(and maybe **ADTs** in general)_ but that allows customization when needed.

Related to #235

### LINKED ISSUE #564: API to allow returning records AND ResultSummary
Currently there's no way to issue a query for records while and accessing the ResultSummary from the same query

BalmungSan:
# ⚠️ BIG REFACTOR ⚠️

In this PR I will propose my new design for **Neotypes** based on explicit decoders rather than the current approach based on typeclasses.

This design is heavily inspired by other libraries like **Skunk** & **Dynosaur**.

### 1. Neo4j data model

The first step of the new design was modeling the data model of **Neo4j** as an ADT.

### 2. Mappers and query core API

Then, we need to reshape the mappers based on the new model.
As well as adapt the query API to be based on explicit values rather than implicit ones.

### 3. Mappers algebra and base instances

The next step is to provide the base instances and algebra that will be used to create customized mappers for advanced use cases.

### 4. Explicit automatic derivation

Finally, we will also provide a way to derive instances of the mappers; but explicitly.
This would mean that end users who don't care about advanced use cases can get a UX similar to the current approach.

## API

I still haven't done proper docs about the new API but you can see how its usage looks like on the unit tests: https://github.com/BalmungSan/neotypes/blob/explicit-decoders/core-test/src/test/scala/neotypes/DriverSpec.scala

-----

# Update

While many important points still remain, like: Fixing the other modules, re-adding again all the tests I deleted, and adding docs.
The fundamental work of this PR is finished with that last push I just made 🎉

I want to once again thank everyone who got involved and if you want to give a final review now that the implementation is over that would be great.

> **Note**: I will need to take a break from the development, since I need to resume and finish the written paper associated with this PR.

-----

_Special thanks to Aly (@s5bug) and Arman (@armanbilge) who helped me polish the design and implementation on the **Scala** **Discord** server_. 😄

-----

This work was the basis of my master's degree final project, the memories of the development can be found here: https://www.dropbox.com/s/tmx1p3y75uzg4ip/Memoria.pdf?dl=0

> They are in Spanish, but at the end, there is a small paper in English summarizing everything.

-----

Closes #236
Fixes #564


geoffjohn11:
I've been looking forward to this PR!

j-mie6:
Looks pretty intuitive from the test cases!

core/src/main/scala/neotypes/model.scala
```diff
@@ -0,0 +1,198 @@
...
+    map.map {
+      case (key, value) =>
+        key -> value.underlying
+    }.asJava
+}
+
+/** Data types supported by Neo4j. */
+object types {
+  /** Parent type of all Neo4j types. */
+  sealed trait NeoType extends Product with Serializable
```
armanbilge, line 30: NIt: I prefer using `sealed abstract class` for ADTs. It enforces single-inheritance, is binary-compatible with Scala 3 enum, and theoretically has slightly better performance :)

core/src/main/scala/neotypes/model.scala
```diff
@@ -0,0 +1,198 @@
...
+
+  /** Represents a Neo4j Node. */
+  final case class Node(
+      elementId: String,
+      labels: Set[String],
+      properties: Map[String, Value]
+  ) extends Entity {
+    /** Checks if this Node contains the given label; case insensitive. */
+    def hasLabel(label: String): Boolean =
+      labels.contains(label.trim.toLowerCase)
```
armanbilge, line 73: Should labels be a case-insensitive string?

https://github.com/typelevel/case-insensitive

core/src/main/scala/neotypes/model.scala
```diff
@@ -0,0 +1,198 @@
...
+
+    final def keys: Set[String] =
+      properties.keySet
+
+    final def values: Iterable[NeoType] =
+      properties.values
+  }
+
+  /** Represents a Neo4j heterogeneous list (composite type) */
+  final case class NeoList(values: Iterable[NeoType]) extends NeoType
```
armanbilge, line 53: Why not a concrete type?

core/src/main/scala/neotypes/model.scala
```diff
@@ -0,0 +1,198 @@
...
+    final case class ListValue[V <: SimpleValue](values: Iterable[V]) extends Value
+    sealed trait SimpleValue extends Value
+    sealed trait NumberValue extends SimpleValue
+    final case class Integer(value: Int) extends NumberValue
+    final case class Decimal(value: Double) extends NumberValue
+    final case class Str(value: String) extends SimpleValue
+    final case class Bool(value: Boolean) extends SimpleValue
+    final case class Bytes(value: ArraySeq[Byte]) extends SimpleValue
+    final case class Point(value: NeoPoint) extends SimpleValue
+    final case class Duration(value: NeoDuration) extends SimpleValue
```
armanbilge, line 151: Why not use a Scala duration here?

core/src/main/scala/neotypes/model.scala
```diff
@@ -0,0 +1,198 @@
...
+  sealed trait Value extends NeoType
+  object Value {
+    final case class ListValue[V <: SimpleValue](values: Iterable[V]) extends Value
+    sealed trait SimpleValue extends Value
+    sealed trait NumberValue extends SimpleValue
+    final case class Integer(value: Int) extends NumberValue
+    final case class Decimal(value: Double) extends NumberValue
+    final case class Str(value: String) extends SimpleValue
+    final case class Bool(value: Boolean) extends SimpleValue
+    final case class Bytes(value: ArraySeq[Byte]) extends SimpleValue
```
armanbilge, line 149: scodec-bits `ByteVector` could be good here. The library has no dependencies and is very stable.

core/src/main/scala/neotypes/model.scala
```diff
@@ -0,0 +1,198 @@
...
+    final case class DateTme(value: JZDateTime) extends TemporalInstantValue
+    final case object NullValue extends SimpleValue
+  }
+}
+
+/** Exceptions provided by this library. */
+object exceptions {
+  sealed abstract class NeotypesException(message: String, cause: Option[Throwable] = None) extends Exception(message, cause.orNull)
+
+  final case object TransactionWasNotCreatedException extends NeotypesException(message = "Couldn't create a transaction")
```
armanbilge, line 166: Seems like these singleton exceptions should extend `NoStackTrace`, since their stack traces will not point to the callsite anyway.

Personally I am a bit skeptical about mixing case-class based ADTs with exceptions. `case class`es have a particular notion of equality that doesn't really take into account stacktraces etc. Even if you throw the same exception with the same message from the same place at two different points in time, are they _really_ equal? 🤔

core/src/main/scala/neotypes/model.scala
j-mie6: Or at least `Seq`, seems like from the description `List` this is supposed to have some actual ordering?

core/src/main/scala/neotypes/model.scala
j-mie6: I disagree here, I think it's useful to have case class exceptions for the ability to pattern match on them in recovery combinators.

armanbilge: You can get unapply without using case classes :) or their specific equality semantics

j-mie6: That's true, I'd +1 _that_.

core/src/main/scala/neotypes/model.scala
BalmungSan: Yeah, sadly **neotypes** has a strong policy of only depending on the **neo4j-java-driver** for its `core` module.
Even **Shapeless** was moved to the `generic` module.

core/src/main/scala/neotypes/model.scala
BalmungSan: Ditto dependencies :)

core/src/main/scala/neotypes/model.scala
BalmungSan: Originally I used `List` but that would require unnecessary copying since most of the time this would be created from a **Java** collection using `asScala`.

It was for that reason that I asked the other day if it was an easy way to get an immutable view over some collection without copying it.

I guess I could create my own `Seq` or even `IndexedSeq` out of it.
But, since usually users won't access this directly but `ResultMapper.values`, and that one will call `Map.values` which also returns `Iterable` I decided it would be better to just leave it like that.

core/src/main/scala/neotypes/model.scala
BalmungSan: Well, for once:

> This class is not meant as a general purpose representation of time, it is optimized for the needs of scala.concurrent.


Also, the **Neo4j** one extends **Java's** [`TemporalAmount`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/time/temporal/TemporalAmount.html) which makes it easy to use; e.g. converting it to a `java.time.Duration`.

Finally, again, this ADT while public is mostly internal.
Ideally, users won't need to reach to it but use the pre-defined mappers, and I can _(and will, thanks for the suggestion)_ add a constant to read data as a **Scala** `FinitieDuration`

core/src/main/scala/neotypes/model.scala
BalmungSan: > Seems like these singleton exceptions should extend NoStackTrace

I thought I did, but it seems I added it to the `ResultMapperException` hierarchy rather than the `NeotypesException`.

> Personally I am a bit skeptical about mixing case-class based ADTs with exceptions. case classes have a particular notion of equality that doesn't really take into account stacktraces etc

Fair.

> You can get unapply without using case classes

Yup :p

core/src/main/scala/neotypes/query.scala
```diff
@@ -1,568 +1,479 @@
...
-    */
-  final def list[F[_], O](driver: Driver[F])
-                         (implicit rm: ResultMapper[T]): F[List[T]] =
-    effectDriverTransaction(driver)(tx => tx.list(query, params))
+/** Represents a query that doesn't produce results when executed. */
+final class ExecuteQuery private[neotypes] (
+    val query: String,
+    val params: Map[String, QueryParam]
+) extends BaseQuery {
+  override def withParams(params: Map[String, QueryParam]): ExecuteQuery =
```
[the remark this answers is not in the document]
BalmungSan: Well, to be totally honest, I am only leaving that because it already was there and I didn't want to break even more things.

But, in practice, I doubt anyone has used this.
Like, is not common to create a **cypher** query with a dummy set of parameters to then replace / change / manipulate them.

core/src/main/scala/neotypes/model.scala
BalmungSan: @armanbilge actually, during the implementation of `Parser` I realized that I needed to produce a concrete collection, so I switched back to `List` 👍

BalmungSan:
For the record, on my PC the tests of the modules `core`, `cats-effect` & `fs2` are green!

zio-stream/src/main/scala/neotypes/zio/stream/ZioStreams.scala
```diff
@@ -1,22 +1,25 @@
...
+import scala.collection.Factory
 
 trait ZioStreams {
   implicit final val zioStream: neotypes.Stream.Aux[ZioStream, Task] =
     new neotypes.Stream[ZioStream] {
       override final type F[T] = Task[T]
 
-      override final def fromPublisher[A](publisher: Publisher[A]): ZioStream[A] =
-        Adapters.publisherToStream(toPublisher(publisher), bufferSize = 16)
+      override final def fromPublisher[A](publisher: => Publisher[A], chunkSize: Int): ZioStream[A] =
```
jmcclell, line 18: ChunkSize < 2 will cause an assertion failure in the internals of Zio as it eventually gets passed directly to Zio's RingBuffer impl. Causes a test failure in this branch. Not sure how you want to handle it so just commenting here to save you the trouble.

BalmungSan: Why I am not surprised of **ZIO** making my life harder than it already is.

There are many singleton `Publishers` in this program and generally in the wild, why does the buffer have to be of size `2`?!

jmcclell: You can see the assertion [here](https://github.com/zio/zio/blob/ba58425b23a4beb60b511ce2bb5babafc40de591/core/js/src/main/scala/zio/internal/RingBuffer.scala#L28). It looks like the code was inspired by [this impl](https://www.1024cores.net/home/lock-free-algorithms/queues/bounded-mpmc-queue), which has the same assertion.

sebastianvoss:
This is a really cool feature. Thanks for putting all the effort! Do you already have a rough idea when you can make a release or RC available?

BalmungSan:
Hi @sebastianvoss thanks for your interest in the project and sorry for taking too long.
The good and sad news is that we are really close, it just has been hard for me to find time.

What is missing is just:

+ Fix the `generic` tests.
+ Re-add all the `core` tests that I deleted.
+ Fix the `cats-data`, `refined`, and `enumeratum` modules
+ Merge `main` and fix conflicts
+ Fix any bug that is found during that time

That should be like 3 at most 4 days of work.
Maybe tomorrow I will be able to make some progress, and if nothing happens next weekend as well.

Thus, I hope to make an RC release in two weeks.

BalmungSan:
I want to apologize to everyone for taking so long to finish this, especially when all that was missing was just tests and fixing the other modules. It was hard to find the motivation on working on those minor details... But it is almost ready!

I just finished fixing the broken modules.
Right now, I am just fixing some errors that happened after the `merge` with `main`. Hopefully, that won't take me too long.
After that, all that remains is fixing the docs, and then we will be able to do a proper _(pre)_ release.
I will work no stop until that happens!

BalmungSan:
Okay, I finished fixing the tests!

But I noticed **AkkaStreams** is broken :(
Also, it seems `execute.void` is swallowing exceptions, but not sure.

After I fix those two things, all that remains is updating the docs.
I really hopped to release this weekend, but at least we are now very close 🙂

BalmungSan:
Well after a whole day of debugging, I fixed **AkkaStreams** _(the morale of the story is that eager computations are bad!)_

I will work on the docs this week, or maybe next weekend.
For now, I am drained of energy.

But, I am hopeful I will do the release before May ends! :)

BalmungSan:
Okay, I think I am finally done! :D

In a couple of days I will start doing releases, stay tuned.

PR #665, opened 2023-08-25
Prepare for scala 3

i10416:
These changes are extracted from https://github.com/neotypes/neotypes/pull/555 just for reducing diffs.

This PR addresses Scala 3 compiler warnings and moves some Scala 2 specific sources to dedicated folder.


core/src/main/scala/neotypes/mappers/ResultMapper.scala
```diff
@@ -551,10 +551,10 @@ object ResultMapper extends ResultMappersLowPriority0 with BoilerplateResultMapp
   object CoproductDiscriminatorStrategy {
 
     /** Discriminates cases based on a label of Node. */
-    final case object NodeLabel extends CoproductDiscriminatorStrategy[String]
```
i10416, line 554: redundant final modifiers

i10416:
I'm going to submit follow-up PRs so that I will not introduce a lot of diffs at once.

build.sbt
```diff
@@ -117,6 +117,29 @@ lazy val root = (project in file("."))
     )
   )
 
+lazy val scalaVersionDependentSettings = Def.settings(
+  libraryDependencies ++= (if (scalaVersion.value.startsWith("2."))
```
BalmungSan, line 121: Could you please start the `if` in the next line so the block doesn't get indented too far right?

build.sbt
```diff
@@ -117,6 +117,29 @@ lazy val root = (project in file("."))
...
+                             COMPILE(
+                               scalaVersion("org.scala-lang" % "scala-reflect" % _).value
+                             )
+                           else Seq.empty),
```
BalmungSan, line 125: I would close the the parenthesis in the next line :)

build.sbt
```diff
@@ -117,6 +117,29 @@ lazy val root = (project in file("."))
...
+  scalacOptions ++= (if (scalaVersion.value.startsWith("2."))
+                       Seq(
+                         "-Ywarn-macros:after",
+
+                         // Ensure we publish an artifact linked to the appropriate Java std library.
+                         "-release:17",
```
BalmungSan, line 131: This flag MUST be always added, so let's move it outside of this block.
Actually, it is already part of the `commonSettings`, so there is no need to re-add it here.

build.sbt
```diff
@@ -117,6 +117,29 @@ lazy val root = (project in file("."))
...
+
+                         // Implicit resolution debug flags.
+                         "-Vimplicits",
+                         "-Vtype-diffs",
+                         // Make all warnings verbose.
+                         "-Wconf:any:warning-verbose"
```
BalmungSan, line 137: I am not very familiar with **Scala 3**, but are you sure those do not exist there?

BalmungSan (commented):
Hi, thanks a lot for taking care of this!
Left a couple of minor style comments and a real observation that should be addressed before merging.

BTW, maybe the **microsite** should remain in **Scala 2** for now? Since it is the lower common denominator.
Also, is there a good **Enumeratum equivalent in **Scala 3**? Maybe just using the `enum` magic will be enough? Otherwise, instead of moving the sources to a **Scala 2** only folder, we need to disable cross-build and publishing for the whole module, since it will be empty for **Scala 3**.

i10416:
Thanks for your reviews.

Actually, some changes are made to carelessly silence warnings. I'will clean up them soon.

> Also, is there a good **Enumeratum equivalent in Scala 3? Maybe just using the enum magic will be enough

From #555, I guessed Scala 3 enum would be eonugh , but perhaps, is not.  Therefore, I will investigate it too.

build.sbt
i10416: It seems they don't and their alternatives do not too.

https://github.com/scala/docs.scala-lang/blob/main/_overviews/scala3-migration/options-lookup.md

BalmungSan:
> From https://github.com/neotypes/neotypes/pull/555, I guessed Scala 3 enum would be eonugh , but perhaps, is not. Therefore, I will investigate it too.

Yeah, I mean if we can provide functionality similar to our own **Enumeratum** module to **Scala 3** `enums` then great, we can and should cross-compile / publish the module.
But if not, the best is to just mark it as a **Scala 2** only module.

build.sbt
i10416: Addressed.

BalmungSan (approved):

build.sbt
i10416: Done 👍

build.sbt
i10416: Done 👍

i10416:
After a quick search, I think we should exclude enumeratum module from Scala 3 build for now because Scala 3 enum is not 100% compatible with enumeratum and enumeratum has difficulty cross-building in Scala 3.

> looks like support for custom enumLabel was https://github.com/lampepfl/dotty/pull/9549, then later https://github.com/lampepfl/dotty/pull/9989, and I can't figure out how to get it working


> So I don't think any further progress can be made until that's fixed.

https://github.com/lloydmeta/enumeratum/issues/300

BalmungSan:
@i10416 This one is ready to merge then, yes?

i10416:
Yes. If you don't have any concern. I believe changes in this PR are just for preparation and therefore won't break anything.

BalmungSan (approved):

BalmungSan:
@i10416 cool, after the CI is green I will merge this and make a minor release since it has been a while and there a couple of important dependencies upgrades.

PR #666, opened 2023-08-25
cross compile core module in Scala 3

i10416:
follow up https://github.com/neotypes/neotypes/pull/665
There are some duplication to separately run tests for Scala 3 like `++3.3.0 core/test`, which will be removed after successful Scala 3 migration.

Now, core module compiles both Scala 2 and Scala 3.

This PR contains CypherStringInterpolator implementation both for Scala 2 and Scala 3.

This PR also contains fix for the Null case:
- impl: https://github.com/neotypes/neotypes/pull/666/files#diff-44d3b4d2acba583ac01bbd2c1df867d66033e58cb75f4c00ae1baaeb2098bb43R32
- test: https://github.com/neotypes/neotypes/pull/666/files#diff-0a69ff6895e8e549ac1ec2bf58ebbb06333d90319a3b3b6bc4e59d011c31f831R223

To check diffs between #665 and #666, see https://github.com/i10416/neotypes/pull/1/files


generic/src/test/scala/neotypes/syntax/CypherQueryInterpolationSpec.scala
```diff
@@ -0,0 +1,114 @@
...
+
+import neotypes.BaseSynchronousSpec
+import neotypes.generic.implicits._
+import neotypes.model.query.QueryParam
+import neotypes.syntax.cypher._
+
+final class CypherGenericQueryInterpolationSpec extends BaseSynchronousSpec {
+  import CypherGenericQueryInterpolationSpec._
+
+  behavior of s"Neotypes Cypher String interpolator"
```
BalmungSan, line 11: Something something

```scala
behavior of s"Neotypes Cypher String interpolator plus generic capabilities"
```

test-helpers/src/test/scala/neotypes/BaseSpec.scala
```diff
@@ -0,0 +1,20 @@
+package neotypes
```
BalmungSan, line 1: Why did this have to be moved outside of `core-tests`?

core/src/main/scala/neotypes/internal/utils.scala
```diff
@@ -34,4 +37,82 @@ private[neotypes] object utils {
 
     loop(acc = factory.newBuilder)
   }
+  def createQuery(queryData: Either[String, QueryArg]*): DeferredQueryBuilder = {
```
BalmungSan, line 40: We should reuse these utils as well on the **Scala 2** version of the macro.

core/src/main/scala-3/neotypes/CypherStringInterpolator.scala
```diff
@@ -0,0 +1,99 @@
...
+    interleave(parts, queries, endWithHashTags, List.empty)
+  private[this] def interleave(
+    parts: List[Boolean => Res],
+    queries: List[Res],
+    endWithHashTags: List[Boolean],
+    acc: List[Res]
+  ): List[Res] =
+    (queries, parts, endWithHashTags) match {
+      case (Nil, l2, b)                   => acc.reverse ++ (l2 zip b).map { case (e, b) => e(b) }
+      case (l1, Nil, _)                   => acc.reverse ++ l1
```
BalmungSan, line 95: Take a look to `reverse_:::`

core/src/main/scala-3/neotypes/CypherStringInterpolator.scala
```diff
@@ -0,0 +1,99 @@
...
+
+  type Res = Either[String, QueryArg]
+
+  def macroImpl(argsExpr: Expr[Seq[Any]], clz: Expr[StringContext])(using quotes: Quotes): Expr[DeferredQueryBuilder] =
+    import quotes.reflect.*
+
+    def invert[T: Type](exprs: Expr[Seq[T]]): List[Expr[T]] =
+      exprs match {
+        case Varargs(props)                => props.toList
+        case '{ Seq(${ Varargs(props) }) } => props.toList.map(_.asExprOf[T])
```
BalmungSan, line 22: It may be worth using `view` here to avoid the double traversal.

core/src/main/scala-3/neotypes/CypherStringInterpolator.scala
```diff
@@ -0,0 +1,99 @@
...
+        case _                             => println(s"Unable to match ${exprs.show} "); null
```
BalmungSan, line 23: I think there is a better way to fail a macro with a compilation error.

BalmungSan (commented):
BTW, please add an empty line between all `defs` / `vals` :)

generic/src/test/scala/neotypes/syntax/CypherQueryInterpolationSpec.scala
```diff
@@ -0,0 +1,114 @@
...
+
+  it should "interpolate literal nulls" in {
+    val (query, params, _) = c"CREATE (u: User { np: ${null} }) RETURN u".build()
+
+    query shouldBe "CREATE (u: User { np: $p1 }) RETURN u"
+    params shouldBe Map(
+      "p1" -> QueryParam.NullValue
+    )
+  }
+}
```
BalmungSan, line 106: These two are duplicated here and in the `core` spec, are they not?

core/src/main/scala-3/neotypes/CypherStringInterpolator.scala
i10416: Yeah, thanks for reminding me of `report.errorAndAbort`.

generic/src/test/scala/neotypes/syntax/CypherQueryInterpolationSpec.scala
i10416: Yes, they're duplicated. I copied them to separately run tests for core module in Scala 3  like `++3.3.0 core/test`.

The change you pointed out in https://github.com/neotypes/neotypes/pull/666#discussion_r1306502360 is introduced for the same reason.

After all the modules are compiled in Scala 3, we don't need duplication.

core/src/main/scala-3/neotypes/CypherStringInterpolator.scala
i10416: Addressed.

core/src/main/scala-3/neotypes/CypherStringInterpolator.scala
i10416: Done.

generic/src/test/scala/neotypes/syntax/CypherQueryInterpolationSpec.scala
i10416: Addressed.

core/src/main/scala/neotypes/internal/utils.scala
i10416: Done.

core/src/main/scala-3/neotypes/CypherStringInterpolator.scala
i10416: Sorry, but I couldn't get what you mean.

```scala
List(1,2,3,4,5).reverse_:::(List(6)) 
res0: List[Int] = List(6, 1, 2, 3, 4, 5)

List(1,2,3,4,5).reverse ++ List(6) 
res1: List[Int] = List(5, 4, 3, 2, 1, 6)
```

test-helpers/src/test/scala/neotypes/BaseSpec.scala
i10416: https://github.com/neotypes/neotypes/pull/666#discussion_r1306543528

i10416:
It seems Scala 2.13 expands macro at call site, which causes compile error due to visibility scope.
I think it does not make much difference to make `createQuery` public and make other methods in utils private because utils is already in `internal` package.
Users would know it is internal and they should use it on their own risk.
 
https://github.com/neotypes/neotypes/pull/666/commits/50f7b78f4d3397281dc5a78cfeed63e8227c6cfd

build.sbt
```diff
@@ -129,21 +129,34 @@ lazy val scalaVersionDependentSettings = Def.settings(
     if (scalaVersion.value.startsWith("2."))
       scalacOptions.value
     else
-      scalacOptions.value.filterNot(Set("-Ywarn-macros:after", "-Vimplicits", "-Vtype-diffs"))
+      scalacOptions
+        .value
+        .filterNot(Set("-Ywarn-macros:after", "-Vimplicits", "-Vtype-diffs", "-Wconf:any:warning-verbose"))
```
i10416, line 134: overlooked "-Wconf:any:warning-verbose" in the previous PR.

core/src/main/scala/neotypes/internal/utils.scala
```diff
@@ -3,16 +3,19 @@ package internal
 
 import scala.collection.Factory
 import scala.collection.mutable.Builder
+import neotypes.query.QueryArg
+import neotypes.model.query.QueryParam
+import neotypes.query.DeferredQueryBuilder
 
-private[neotypes] object utils {
+object utils {
```
i10416, line 10: See my comment: https://github.com/neotypes/neotypes/pull/666#issuecomment-1694589087

i10416:
Strangely, test on graalvm takes a quite long.

i10416:
@BalmungSan Could you run tests again?
Tests in #673, which based on this PR, passes. Thus, changes in this PR should pass too.

core/src/main/scala-3/neotypes/CypherStringInterpolator.scala
BalmungSan: @i10416 remove the dot :)
Is a method that ends in `:` thus, it is being called on `List(6)`

test-helpers/src/test/scala/neotypes/BaseSpec.scala
BalmungSan: @i10416 so if I understand correctly, this will also be removed once everything cross-compiles?

build.sbt
BalmungSan: That should work on **Scala 3**, shouldn't it?

build.sbt
```diff
@@ -153,8 +166,10 @@ lazy val generic = (project in file("generic"))
     libraryDependencies ++=
       COMPILE(
         "com.chuusai" %% "shapeless" % shapelessVersion
-      )
+      ),
+    Test / scalacOptions += "-Wconf:cat=other-pure-statement&msg=org.scalatest.Assertion:s"
```
BalmungSan, line 170: This should be on the common settings since we want this exception on all tests.

core/src/main/scala-3/neotypes/CypherStringInterpolator.scala
```diff
@@ -0,0 +1,129 @@
...
+    */
+  def apply(interpolations: List[Boolean => Res], stringFlagments: List[Either[String, String]]): List[Res] =
+    interleave(interpolations, stringFlagments, List.empty)
+  private[this] def interleave(
+    interpolations: List[Boolean => Res],
+    stringFlagments: List[Either[String, String]],
+    acc: List[Res]
+  ): List[Res] =
+    (stringFlagments, interpolations) match
+      case (Nil, Nil) => acc
```
BalmungSan, line 113: For consistency, always break line after the `=>`
Also, please leave an empty line with the next `case`

core/src/main/scala-3/neotypes/CypherStringInterpolator.scala
```diff
@@ -0,0 +1,129 @@
...
+      // this should be always query interpolations because if there is a plain value interpolation, there should be `#` prefix.
```
BalmungSan, line 114: Please move the comment under the `case` block for clarity.

core/src/main/scala-3/neotypes/CypherStringInterpolator.scala
```diff
@@ -0,0 +1,129 @@
...
+        interleave(interps, tail, interp(true) :: Left(nthPartExpectQueryValueInsert) :: acc)
+      case (Right(theLastPart) :: Nil, Nil) => (Left(theLastPart) :: acc).reverse
+      case (Left(theLastPartIsAHashTag) :: Nil, Nil) =>
+        (Left(theLastPartIsAHashTag) :: acc).reverse
+      case (Right(partExpectQueryValueInsert) :: _, Nil) =>
+        // this should be unreachable
+        throw new Exception(s"interpolation value number mismatch")
+      case (Left(partExpectPlainValueInsert) :: _, Nil) =>
+        // this should be unreachable
+        throw new Exception(s"interpolation value number mismatch")
```
BalmungSan, line 128: Maybe just use `case _ =>` here to catch both cases.
Also, couldn't we use here the same `abortMacro` thing?

test-helpers/src/test/scala/neotypes/BaseSpec.scala
i10416: Yes.

core/src/main/scala-3/neotypes/CypherStringInterpolator.scala
i10416: `report` APIs are in `quotes.reflect` package and  you need `Quotes` in scope, so I think it is not appropriate here.

build.sbt
i10416: Ah, it found it works after checking again.

i10416: Thanks.

build.sbt
i10416: I moved it to commonSettings.

core/src/main/scala-3/neotypes/CypherStringInterpolator.scala
i10416: Done.

core/src/main/scala-3/neotypes/CypherStringInterpolator.scala
i10416: Done.

core/src/main/scala-3/neotypes/CypherStringInterpolator.scala
BalmungSan: Okay, but then how does this exception interact with the macro? Or will this get thrown during runtime?
It doesn't matter since it is impossible, I just want to understand better how this works :)

i10416: >  I just want to understand better how this works :)

When you throw an exception inside `inline def aMethodThatUsesQuoteAPI`, the exception is thrown at macro expansion.

For example, if I put a line that throw an exception(see `macroImpl` method),

```scala
final case class CypherStringInterpolator(private val sc: StringContext) extends AnyVal {
  inline def c(inline args: Any*): DeferredQueryBuilder = ${ CypherStringInterpolator.macroImpl('args, 'sc) }
}

object CypherStringInterpolator {

  type Res = Either[String, QueryArg]

  def macroImpl(argsExpr: Expr[Seq[Any]], clz: Expr[StringContext])(using quotes: Quotes): Expr[DeferredQueryBuilder] =
    throw new Exception("boom") // <--- exception here
    import quotes.reflect.*

    def invert[T: Type](exprs: Expr[Seq[T]]): List[Expr[T]] =
      exprs match {
        case Varargs(props)                => props.toList
        case '{ Seq(${ Varargs(props) }) } => props.view.map(_.asExprOf[T]).toList
        case _                             => report.errorAndAbort(s"Unable to match ${exprs.show}")
      }

```
And then I run `++3.3.0 core/test:compile`
  I got the following.

```
sbt:root> ++3.3.0 core/test:compile
[info] Setting Scala version to 3.3.0 on 2 projects.
[info] Excluded 14 projects, run ++ 3.3.0 -v for more details.
[info] Reapplying settings...
[info] set current project to root (in build file:/path/to/neotypes/)
[warn] sbt 0.13 shell syntax is deprecated; use slash syntax instead: core / Test / compile
[info] compiling 1 Scala source to /path/to/neotypes/core/target/scala-3.3.0/test-classes ...
[error] -- Error: /path/to/neotypes/core/src/test/scala/neotypes/syntax/CypherQueryInterpolationSpec.scala:12:29 
[error] 12 |    val (query, params, _) = c"CREATE (a: Test { name: 'John' })".build()
[error]    |Exception occurred while executing macro expansion.
[error]    |java.lang.Exception: boom
[error]    |    at neotypes.query.CypherStringInterpolator$.macroImpl(CypherStringInterpolator.scala:17)
[error]    |
```

On the other hand, when I throw an exception outside a quote scope, the exception is thrown at runtime.

For example

```scala
  def apply(interpolations: List[Boolean => Res], stringFlagments: List[Either[String, String]]): List[Res] =
    throw new Exception("boom") <--- exception here
    interleave(interpolations, stringFlagments, List.empty)
```

and `++3.3.0 core/test:compile` is quiet, but `++3.3.0 core/test` throws the exception at runtime.

core/src/main/scala-3/neotypes/CypherStringInterpolator.scala
i10416: Oh... a bit surprising result...

BalmungSan (approved):
Cool thanks for all the changes and explanations.
This one is ready as well, right?

i10416:
Yes, thanks for reviews!

Issue #469, opened 2022-01-29
Request: Scala 3 support 


oltyant:
When can neotypes support Scala 3?

BalmungSan:
Hi @oltyant the plan was to first release a version supporting **CE3** and then **Scala 3**.
But **Monix** has kept us blocked with the first thing.

Anyways, I hope to find some time soon to give the **Scala 3** bit a go.

oltyant:
Hi @BalmungSan , thanks for the quick answer!

BalmungSan:
# Roadmap

- [x] Move `generic` to its own module and fix the tests - #548
- [ ] Add a **Scala 3** version of `generic` module - _(WIP)_ #555
- [ ] Add a **Scala 3** version of **cypher** string interpolator - _(WIP)_ #555
- [ ] Cross-compile all other modules _(this should work out of the box)_

PerWiklander:
Any updates on this? Is it included in the upcoming 1.0 release?

BalmungSan:
@PerWiklander hey, sorry, forgot to close this one. Actually the latest `RC` of **neotypes** fully supports **Scala 3**, this was released almost a year ago.

I haven't made a `1.0.0` proper release because I wanted to change the way the `sc` string interpolator worked, but I haven't found the time nor the motivation to do it.

Just use the latest `RC`, is pretty stable and I probably will just promote that one as final `1.0.0` this weekend.

BalmungSan:
https://github.com/neotypes/neotypes/releases/tag/v1.0.0

PR #667, opened 2023-08-25
cross compile generic module in Scala 3

i10416:
follow up #666 
https://github.com/neotypes/neotypes/issues/469
This PR intends to build core and generic modules in Scala 3. ~Cross build of other extra modules(cats, zio, akka, etc.) will be done in upcoming PR(s)~

Diffs from separate-tests-for-incremental-migration: https://github.com/i10416/neotypes/pull/2/files

- [x] derive product result map
- [x] derive coproduct result map
- [x] derive query params


generic/src/test/scala/neotypes/syntax/CypherQueryInterpolationSpec.scala
```diff
@@ -0,0 +1,115 @@
+package neotypes.syntax
+
+import neotypes.BaseSynchronousSpec
+import neotypes.generic.implicits._
+import neotypes.generic.implicits.given
```
i10416, line 5: I'm not sure whether Scala 2 allows this `neotypes.generic.implicits.given` instead of given with backticks  on purpose or by mistake.

build.sbt
```diff
@@ -117,18 +117,54 @@ lazy val root = (project in file("."))
...
+    crossScalaVersions := Seq("2.13.11", "3.3.0"),
     Compile / sourceGenerators += Boilerplate.generatorTask.taskValue,
     libraryDependencies ++=
       PROVIDED(
         "org.neo4j.driver" % "neo4j-java-driver" % neo4jDriverVersion
-      ) ++ COMPILE(
-        scalaVersion("org.scala-lang" % "scala-reflect" % _).value
-      )
+      ),
+    Test / scalacOptions += "-Wconf:cat=other-pure-statement&msg=org.scalatest.Assertion:s"
```
i10416, line 164: This is necessary to use this hack https://github.com/neotypes/neotypes/pull/667/files#r1305977657

build.sbt
```diff
@@ -138,8 +174,11 @@ lazy val generic = (project in file("generic"))
     libraryDependencies ++=
       COMPILE(
         "com.chuusai" %% "shapeless" % shapelessVersion
-      )
+      ),
+    Test / scalacOptions += "-Wconf:cat=other-pure-statement&msg=org.scalatest.Assertion:s"
   )
+  .dependsOn(`test-helpers` % "test->test")
+  .settings(scalacOptions += "-Wconf:origin=neotypes.generic.implicits.given:s")
```
i10416, line 181: This is necessary to use this hack https://github.com/neotypes/neotypes/pull/667/files#r1305977657

i10416:
Note for myself.

```scala
//> using scala "3.3.0"
//> using dep "org.typelevel::cats-core::2.9.0"
//> using dep "io.github.neotypes::neotypes-core::1.0.0-M2-SNAPSHOT"
//> using dep "org.typelevel::shapeless3-deriving::3.1.0"
//> using dep "org.neo4j.driver:neo4j-java-driver:5.11.0"
package neotypes.generic
import shapeless3.deriving.*
import neotypes.mappers.ResultMapper
import neotypes.syntax.all.*
import neotypes.mappers.ResultMapper.DerivedProductMap
import neotypes.model.types.NeoObject
import neotypes.model.exceptions.ResultMapperException
import neotypes.model.exceptions.IncoercibleException

trait CaseClassDerivedProductMap[P <: Product]
    extends ResultMapper.DerivedProductMap[P]
object CaseClassDerivedProductMap:

  type Acc = (Int, NeoObject, Option[ResultMapperException])
  implicit final def derivedProductInstance[P <: Product](implicit
      labelled: => Labelling[P],
      inst: => K0.ProductInstances[ResultMapper, P]
  ): ResultMapper.DerivedProductMap[P] = new CaseClassDerivedProductMap[P]:
    override def map(obj: NeoObject): Either[ResultMapperException, P] =
      val ((_, _, decodeOpt), resultOpt) = inst.unfold[Acc]((0, obj, None)):
        [t] =>
          (acc: Acc, mapper: ResultMapper[t]) =>
            val (idx, remaining, optErr) = acc
            val fieldName = labelled.elemLabels(idx)
            val decodedHead = obj.getAs(fieldName, mapper)
            (
              (idx + 1, obj, decodedHead.left.toOption),
              decodedHead.toOption
          )
      resultOpt.toRight(
        decodeOpt.getOrElse(
          IncoercibleException(
            "Unexpected exception occurred during derivation.",
            None
          )
        )
      )

object implicits:
  implicit final def deriveCaseClassProductMap[P <: Product](using
      mapper: CaseClassDerivedProductMap[P]
  ): DerivedProductMap[P] =
    mapper

case class A(i: Int, s: String, b: B)
case class B(n: Int, l: Int)
object B:
  import implicits.*

@main
def program =
  import implicits.given
  val bInst:DerivedProductMap[B] = summon[DerivedProductMap[B]]
  val data = NeoMap(Map("n"-> neotypes.model.types.Value.Integer(1),"l"->  neotypes.model.types.Value.Integer(2)))
  def a:ResultMapper[B] = ResultMapper.productDerive[B](bInst.asInstanceOf)  // this compiles, but
  def a:ResultMapper[B] = ResultMapper.productDerive[B](bInst)
  // this fails due to 
  // Found:    (bInst : neotypes.mappers.ResultMapper.DerivedProductMap[neotypes.generic.B]), 
  // Required: ResultMappersLowPriority3.this.DerivedProductMap[neotypes.generic.B]
  println(a.decode(data))
```

`productDerive` is implemented in the following way. `self:  ResultMapper.type` seems culprit. 

```scala
private[mappers] sealed abstract class ResultMappersLowPriority3 { self: ResultMapper.type =>
   ...
    implicit final def productDerive[P <: Product](implicit ev: DerivedProductMap[P]): ResultMapper[P] =
    neoObject.emap(ev.map)

```

i10416:
```diff
- implicit final def productDerive[P <: Product](implicit ev: DerivedProductMap[P]): ResultMapper[P] =
+ implicit final def productDerive[P <: Product](implicit ev: ResultMapper.DerivedProductMap[P]): ResultMapper[P] =
```

As expected, this solves type mismatch.

generic/src/main/scala-2/neotypes/generic/implicits.scala
```diff
@@ -6,7 +6,7 @@ import query.QueryArgMapper.DerivedQueryParams
 
 import shapeless.Lazy
 
-object implicits {
+package object implicits {
```
BalmungSan, line 9: Let's not use `package object` unless we really need to :)

BalmungSan (commented):
I will try to take a look at the `given` stuff, because I think we could and should just enable tasty reader and the compatibility mode in **Scala 2** to be able to import using `given`
Otherwise, we can just resort to keep using `implicits` in **Scala 3**.

generic/src/test/scala/neotypes/syntax/CypherGenericQueryInterpolationSpec.scala
```diff
@@ -81,28 +81,6 @@ final class CypherGenericQueryInterpolationSpec extends BaseSynchronousSpec {
       "p3" -> QueryParam(extraProp)
     )
   }
-
```
i10416, line 84: remove duplication

generic/src/test/scala/neotypes/syntax/CypherQueryInterpolationSpec.scala
i10416: I found `import neotypes.generic.implicits._` works fine without `given`.

build.sbt
i10416: No longer necessary.

i10416:
Now, product, coproduct and query params derivation work. I'll clean up duplication and tweak build.

tests/src/test/scala-2/neotypes/enumeratum/EnumeratumSpec.scala
i10416, line 1: enumeratum is not supported in Scala 3（yet）.

BalmungSan (commented):
Thanks for all the hard work @i10416 
I will try to find some time to explore the `given` vs `implicits` issue.

core/src/main/scala/neotypes/mappers/ResultMapper.scala
```diff
@@ -712,7 +712,7 @@ private[mappers] sealed abstract class ResultMappersLowPriority3 { self: ResultM
     * @note
     *   if you need customization of the decoding logic, please refer to the `product` and `productNamed` factories.
     */
-  implicit final def productDerive[P <: Product](implicit ev: DerivedProductMap[P]): ResultMapper[P] =
+  implicit final def productDerive[P <: Product](implicit ev: ResultMapper.DerivedProductMap[P]): ResultMapper[P] =
```
i10416, line 715: Lessen type from ResultMapper.this.DerivedProductMap

tests/src/test/scala-2/neotypes/AsyncSuite.scala
```diff
@@ -0,0 +1,33 @@
...
+      new AsyncDriverSpec(testkit),
+      new AsyncTransactionSpec(testkit),
+      new AsyncDriverTransactSpec(testkit),
+      new AsyncDriverConcurrentUsageSpec(testkit),
+      new AsyncParameterSpec(testkit),
+      new AsyncAlgorithmSpec(testkit),
+      new cats.data.AsyncCatsDataSpec(testkit),
+      new enumeratum.AsyncEnumeratumSpec(testkit),
+      new refined.AsyncRefinedSpec(testkit)
+    )
```
i10416, line 18: Scala 2 modules support enumeratum and refined.

tests/src/test/scala-3/neotypes/AsyncSuite.scala
```diff
@@ -0,0 +1,29 @@
...
+/** Group all the Async specs into one big suite, which can be called for each Async type. */
+abstract class AsyncSuite[F[_]](testkit: AsyncTestkit[F])
+    extends Suites(
+      new AsyncGuaranteeSpec(testkit),
+      new AsyncDriverSpec(testkit),
+      new AsyncTransactionSpec(testkit),
+      new AsyncDriverTransactSpec(testkit),
+      new AsyncDriverConcurrentUsageSpec(testkit),
+      new AsyncParameterSpec(testkit),
+      new AsyncAlgorithmSpec(testkit),
```
i10416, line 14: refined and enumeratum modules are not supported.

tests/src/test/scala-3/neotypes/AsyncSuite.scala
```diff
@@ -0,0 +1,29 @@
...
+      new cats.data.AsyncCatsDataSpec(testkit)
+    )
+
+/** Group all the Stream specs into one big suite, which can be called for each Stream type. */
+abstract class StreamSuite[S[_], F[_], A](testkit: StreamTestkit[S, F, A])
```
i10416, line 19: Provide concrete type.
see https://users.scala-lang.org/t/solved-unreducible-application-of-higher-kinded-type-to-wildcard-arguments/7257

tests/src/test/scala/neotypes/AsyncGuaranteeSpec.scala
```diff
@@ -134,12 +134,12 @@ final class AsyncGuaranteeSpec[F[_]](testkit: AsyncTestkit[F]) extends BaseAsync
 }
 
 object AsyncGuaranteeSpec {
-  final case object InputException extends Exception("Input failed")
```
i10416, line 137: address compiler warning.

tests/src/test/scala/neotypes/AsyncSuite.scala
```diff
@@ -1,12 +1,10 @@
 package neotypes
 
-import org.scalatest.Suites
-
 import scala.concurrent.{ExecutionContext, Future}
 import scala.reflect.ClassTag
 
 /** Testkit used to write specs abstracted from any concrete Async type. */
-abstract class AsyncTestkit[F[_]](implicit ct: ClassTag[F[_]]) {
+abstract class AsyncTestkit[F[_]](implicit ct: ClassTag[F[AnyRef]]) {
```
i10416, line 7: To be honest, I'm not sure this is 100% safe.

tests/src/test/scala/neotypes/StreamSuite.scala
```diff
@@ -1,12 +1,10 @@
 package neotypes
 
-import org.scalatest.Suites
-
 import scala.concurrent.{ExecutionContext, Future}
 import scala.reflect.ClassTag
 
 /** Testkit used to write specs abstracted from any concrete Stream type. */
-abstract class StreamTestkit[S[_], F[_]](val asyncTestkit: AsyncTestkit[F])(implicit ctS: ClassTag[S[_]]) {
+abstract class StreamTestkit[S[_], F[_], A](val asyncTestkit: AsyncTestkit[F])(implicit ctS: ClassTag[S[A]]) {
```
i10416, line 7: To be honest, I'm not sure this is 100% safe too.

tests/src/test/scala/neotypes/StreamSuite.scala
```diff
@@ -29,16 +27,16 @@ abstract class BaseStreamSpec[S[_], F[_]](streamTestkit: StreamTestkit[S, F])
   protected implicit final val S: Stream.Aux[S, F] =
     behaviour.streamInstance
 
-  protected final def streamConcurrently[A](stream1: S[Unit], stream2: S[Unit]): S[Unit] =
+  protected final def streamConcurrently(stream1: S[Unit], stream2: S[Unit]): S[Unit] =
     behaviour.streamConcurrently(stream1, stream2)
 
-  protected final def streamToFList[A](stream: S[A]): F[List[A]] =
+  protected final def streamToFList[B](stream: S[B]): F[List[B]] =
```
i10416, line 33: avoid type parameter shadowing.

tests/src/test/scala/neotypes/AsyncSuite.scala
i10416: scattering type params means this approach. https://github.com/neotypes/neotypes/pull/667#discussion_r1307607994

BalmungSan:
@i10416 

> I think we can just use implicits and implicits works better with Scala 2 than given.

Yeah, I think that is also what other cross-compile libraries like **cats** are doing.
But I want to confirm.

BTW, please remember to merge with `main` and fix the conflict, and then let me know when you think this is ready for merge.

tests/src/test/scala/neotypes/AsyncSuite.scala
```diff
@@ -1,12 +1,10 @@
 package neotypes
 
-import org.scalatest.Suites
-
 import scala.concurrent.{ExecutionContext, Future}
 import scala.reflect.ClassTag
 
 /** Testkit used to write specs abstracted from any concrete Async type. */
-abstract class AsyncTestkit[F[_]](implicit ct: ClassTag[F[_]]) {
+abstract class AsyncTestkit[F[_]](implicit ct: ClassTag[F[Any]]) {
```
i10416, line 7: `ClassTag[F[_]]` cares only about outer `F` and I think `Any` is more precise version of wildcard.

tests/src/test/scala/neotypes/AsyncSuite.scala
```diff
@@ -59,18 +57,3 @@ abstract class AsyncDriverProvider[F[_]](testkit: AsyncTestkit[F])
   override protected final def transact[T](driver: DriverType)(txF: TransactionType => F[T]): F[T] =
     driver.transact(txF)
 }
-
-/** Group all the Async specs into one big suite, which can be called for each Async type. */
-abstract class AsyncSuite[F[_]](testkit: AsyncTestkit[F])
```
i10416, line 64: Some modules are not yet supported in Scala 3, so I added dedicated AsyncSuite for Scala 2 and Scala 3.

tests/src/test/scala/neotypes/BaseIntegrationSpec.scala
```diff
@@ -13,7 +13,7 @@ import scala.concurrent.duration._
 import scala.jdk.DurationConverters._
 
 /** Marker trait to signal that the test needs a driver. */
-trait DriverProvider[F[_]] extends { self: BaseIntegrationSpec[F] =>
+trait DriverProvider[F[_]] extends AnyRef { self: BaseIntegrationSpec[F] =>
```
i10416, line 16: Scala 3 doesn't allow `extends` without something class-like.

tests/src/test/scala/neotypes/StreamSuite.scala
```diff
@@ -1,12 +1,10 @@
 package neotypes
 
-import org.scalatest.Suites
-
 import scala.concurrent.{ExecutionContext, Future}
 import scala.reflect.ClassTag
 
 /** Testkit used to write specs abstracted from any concrete Stream type. */
-abstract class StreamTestkit[S[_], F[_]](val asyncTestkit: AsyncTestkit[F])(implicit ctS: ClassTag[S[_]]) {
+abstract class StreamTestkit[S[_], F[_]](val asyncTestkit: AsyncTestkit[F])(implicit ctS: ClassTag[S[Any]]) {
```
i10416, line 7: same as https://github.com/neotypes/neotypes/pull/667/files#r1308560147

build.sbt
```diff
@@ -179,7 +168,6 @@ lazy val catsEffect = (project in file("cats-effect"))
       "org.typelevel" %% "cats-effect" % catsEffect3Version
     )
   )
-  .settings(scala3CrossCompileSettings)
```
i10416, line 182: no longer necessary as I put shared settings into commonSettings.

build.sbt
```diff
@@ -117,33 +127,6 @@ lazy val root = (project in file("."))
...
-)
-
-lazy val `test-helpers` = (project in file("test-helpers"))
-  .settings(
-    crossScalaVersions := Seq("2.13.11", "3.3.0"),
-    libraryDependencies ++= TEST(
-      "org.scalatest" %% "scalatest" % scalaTestVersion
-    )
-  )
-
```
i10416, line 146: no longer necessary as I put shared settings into commonSettings.

build.sbt
```diff
@@ -152,22 +135,28 @@ lazy val core = (project in file("core"))
     libraryDependencies ++=
       PROVIDED(
         "org.neo4j.driver" % "neo4j-java-driver" % neo4jDriverVersion
-      )
+      ),
+    libraryDependencies ++= (
+      if (scalaVersion.value.startsWith("2."))
+        COMPILE(
+          scalaVersion("org.scala-lang" % "scala-reflect" % _).value
```
i10416, line 142: for string interpolation macros

i10416:
@BalmungSan Ready for review.

core/src/main/scala-3/neotypes/CypherStringInterpolator.scala
```diff
@@ -54,7 +54,14 @@ object CypherStringInterpolator {
...
               $qarg
-            else Left($expr.toString)
+            else
+              // This `asInstanceOf` is necessary because when some class overrides `toString()` without parentheses,
+              // this line causes compile error during macro expansion. Scala distinguishes `toString` from `toString()`,
+              // the former is `Select(tree,"toString")` and the latter is `Apply(tree, "toString", typeArgs = Nil, args = Nil)`.
+              //
+              // For example, `cats.data.Chain` overrides to string `override def toString: String = show(Show.fromToString)`.
+              // Without this `asInstanceOf`, the test for Chain in BaseCatsDataSpec won't compile.
+              Left($expr.asInstanceOf[Object].toString())
```
BalmungSan, line 64: Wonderfull @.@

generic/src/main/scala-3/neotypes/generic/deriveCoproductResultMap.scala
```diff
@@ -0,0 +1,22 @@
...
+    labelled: Labelling[C],
+    inst: K0.CoproductInstances[ResultMapper, C]
+  ): SealedTraitDerivedCoproductInstances[C] with
+    override def options: List[(String, ResultMapper[C])] =
+      labelled
+        .elemLabels
+        .zipWithIndex
+        .map[(String, ResultMapper[C])] { (s, idx) =>
+          // TODO is there a way to not use asInstanceOf here? widen as it is currently implemented will
+          // not suffice
```
BalmungSan, line 19: Why `widen` does not suffice?

generic/src/main/scala-3/neotypes/generic/deriveProductMap.scala
```diff
@@ -0,0 +1,41 @@
...
+  private type Acc = (Int, NeoObject, Option[ResultMapperException])
+  given [P <: Product](using
+    labelled: Labelling[P],
+    inst: K0.ProductInstances[ResultMapper, P]
+  ): CaseClassDerivedProductMap[P] with
+    override def map(obj: NeoObject): Either[ResultMapperException, P] =
+      // `unfold` traverses over Product field types.
+      // For example, `inst.unfold` for `case class A(i:Int, s: String)`
+      // traverses over `Int` and `String`
+      val ((_, _, decodeOpt), maybeDerived) = inst.unfold[Acc]((0, obj, None)):
```
BalmungSan, line 23: It seems you don't need to thread `obj` on the `unfold`

generic/src/main/scala-3/neotypes/generic/deriveProductMap.scala
```diff
@@ -0,0 +1,41 @@
...
+        [t] =>
+          (acc: Acc, mapper: ResultMapper[t]) =>
+            val (idx, remaining, optErr) = acc
+            val fieldName = labelled.elemLabels(idx)
+            val decodedHead = obj.getAs(fieldName, mapper)
+            (
+              (idx + 1, obj, decodedHead.left.toOption),
```
BalmungSan, line 30: I don't understand what is `optErr`, why do we get the `left` here?

generic/src/main/scala-3/neotypes/generic/deriveQueryParams.scala
```diff
@@ -0,0 +1,26 @@
...
+    labelled: Labelling[P],
+    inst: K0.ProductInstances[ParameterMapper, P]
+  ): CaseClassDerivedQueryParams[P] with
+    override def getParams(a: P): List[(String, QueryParam)] =
+      val (m, _) = inst.foldLeft[Acc](a)(List.empty -> 0):
+        [t] =>
+          (acc: Acc, pm: ParameterMapper[t], x: t) =>
+            val (oldM, idx) = acc
+            val field = labelled.elemLabels(idx)
+            val nm = oldM :+ (field -> pm.toQueryParam(x))
```
BalmungSan, line 24: We should prepend rather than append and either use `foldRight` or `reverse` at the end.

tests/src/test/scala/neotypes/BaseIntegrationSpec.scala
BalmungSan: Oh, that `extends` was wrong all along, just remove it.

BalmungSan (commented):
Is there no better way to handle the `ClassTag` thing? We only use that for the name of the class, so maybe we could just use **typename** from tpolecat?

core/src/main/scala-3/neotypes/CypherStringInterpolator.scala
i10416: While minimizing this case, I found a solution that avoids asInstanceOf. It is late at night in my timezone, so I'll work on it tommorow or later.

 
https://discord.com/channels/632150470000902164/875868146949554207/1146122590037999726

generic/src/main/scala-3/neotypes/generic/deriveProductMap.scala
i10416: Yeah, right.

core/src/main/scala-3/neotypes/CypherStringInterpolator.scala
i10416: done

generic/src/main/scala-3/neotypes/generic/deriveProductMap.scala
i10416: some variable names are renamed and the control structure is a bit rewritten for clarity.

i10416: The result of unfold is either a product type without error or the very first decode error and other cases should not happen.

https://github.com/neotypes/neotypes/pull/667/files#diff-07e5571fdd32b0141267ce8f7eb5a3bd0d726901fbe97b5b127b54277fe767deR41-R54

generic/src/main/scala-3/neotypes/generic/deriveCoproductResultMap.scala
i10416: If I write `s -> inst.inject(idx) { [t] => (x: ResultMapper[t]) => x.widen[C] }` it raises `Type argument C does not conform to lower bound t`. It seems `C` does not know/could not infer its relationship with `t`.

generic/src/main/scala-3/neotypes/generic/deriveQueryParams.scala
```diff
@@ -0,0 +1,26 @@
...
+trait CaseClassDerivedQueryParams[P <: Product] extends DerivedQueryParams[P]
+
+object CaseClassDerivedQueryParams:
+  private type Acc = (List[(String, QueryParam)], Int)
+
+  given [P <: Product](using
+    labelled: Labelling[P],
+    inst: K0.ProductInstances[ParameterMapper, P]
+  ): CaseClassDerivedQueryParams[P] with
+    override def getParams(a: P): List[(String, QueryParam)] =
```
i10416, line 18: use `::` + `reverse`

generic/src/main/scala-3/neotypes/generic/deriveProductMap.scala
```diff
@@ -11,31 +11,44 @@ import shapeless3.deriving.*
 
 trait CaseClassDerivedProductMap[P <: Product] extends DerivedProductMap[P]
 object CaseClassDerivedProductMap:
-  private type Acc = (Int, NeoObject, Option[ResultMapperException])
+  private type Acc = (Int, Option[ResultMapperException])
   given [P <: Product](using
     labelled: Labelling[P],
     inst: K0.ProductInstances[ResultMapper, P]
   ): CaseClassDerivedProductMap[P] with
     override def map(obj: NeoObject): Either[ResultMapperException, P] =
```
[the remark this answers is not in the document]
i10416: Done.
By the way, I couldn't find `zip` in `Labelling` class, so I just use `inst.constructA`.

generic/src/main/scala-3/neotypes/generic/deriveProductMap.scala
i10416: solved by another approach https://github.com/neotypes/neotypes/pull/667#discussion_r1310603282

generic/src/main/scala-3/neotypes/generic/deriveCoproductResultMap.scala
i10416: Ah, I found I can inform type bounds here.

https://github.com/typelevel/shapeless-3/blob/4225b010b3fe27b94f194e54cc667ee76f133788/modules/deriving/src/test/scala/shapeless3/deriving/type-classes.scala#L638

BalmungSan (approved):
Cool thanks for everything! :tada:
I will take a closer look during the weekend :)

generic/src/main/scala-3/neotypes/generic/deriveCoproductResultMap.scala
```diff
@@ -0,0 +1,20 @@
...
+  given [C](using
+    labelled: Labelling[C],
+    inst: K0.CoproductInstances[ResultMapper, C]
+  ): SealedTraitDerivedCoproductInstances[C] with
+    override def options: List[(String, ResultMapper[C])] =
+      labelled
+        .elemLabels
+        .zipWithIndex
+        .map[(String, ResultMapper[C])] { (s, idx) =>
+          s -> inst.inject(idx) { [t <: C] => (x: ResultMapper[t]) => x.widen[C] }
```
i10416, line 18: specified a type bound here.

BalmungSan (approved):

BalmungSan (approved):

BalmungSan (approved):

BalmungSan:
@i10416 thanks a lot for the changes! everything looks great!
I will merge as soon as the CI is green.

It seems both **Enumeratum** and **Refined** already support **Scala 3**, I will look into that next weekend, but if you open a new PR before that I won't be mad :p
After we cross-compile those two modules as well _(or at least, give it a chance)_ I will make a new release! :tada:

i10416:
Just for your info: Refined is cross-compiled to Scala 3, but it has a bit different API. Most notably, it does not auto-refine a value into refined-types. See https://github.com/fthomas/refined/blob/4be7f6763c924599f54bae12e36b65c1fff60a26/modules/core/shared/src/main/scala-3.0-/eu/timepit/refined/auto.scala vs https://github.com/fthomas/refined/blob/4be7f6763c924599f54bae12e36b65c1fff60a26/modules/core/shared/src/main/scala-3.0%2B/eu/timepit/refined/auto.scala
It seems you must explicitly lift a value into a refined type value using `refineV` and there's no `refineMV` in Scala 3(yet?).

PR #680, opened 2023-09-04
feat: cross compile enumeratum module in Scala 3

i10416:


BalmungSan (dismissed):

BalmungSan:
We need to also add the spec to the **Scala 3** `AsyncSuite`: https://github.com/i10416/neotypes/blob/feat/enumeratum-module-scala-3/tests/src/test/scala-3/neotypes/AsyncSuite.scala

BalmungSan (approved):

PR #808, opened 2024-10-18
Fix ResultMapper on ordered tuple

### LINKED ISSUE #806: ResultMapper.tuple fails with complex types
This fails:

```scala
  it should "derive a tuple composed of more complex types" in {
    val mapper = ResultMapper.tuple[Option[Int], Either[String, Boolean], List[Int], Map[String, String], Option[Set[Long]]]

    val input = NeoMap(
      properties = Map(
        "opt" -> Value.Integer(5),
        "either" -> Value.Bool(true),
        "list" -> Value.ListValue(List(Value.Integer(0), Value.Integer(10))),
        "map" -> NeoMap(Map("foo" -> Value.Str("bar"), "baz" -> Value.Str("quax"))),
        "none" -> Value.NullValue // it does not fail without this last one
      )
    )
    val result = mapper.decode(input)

    result.value shouldBe (Some(5), Right(true), List(0, 10), Map("foo" -> "bar", "baz" -> "quax"), None)
  }
```

All elements of the tuple are mixed up, because calling `values` to iterate over them calls `properties.valuesIterator.toList`, `properties` being a `Map` that does not keep entries order.

The easiest (but maybe not the right one?) way to fix this is to use a `SeqMap` for `NeoMap.properties` to keep the order of values in the Map, or introduce a distinct:

```scala
/** Represents a Neo4j heterogeneous ordered tuple */
final case class Tuple(properties: SeqMap[String, NeoType]) extends NeoObject
```

turb:
Fixes #806 

Two commits:
- create failing tests
- one possible fix using a `SeqMap`

### DIFF
--- a/tests/src/test/scala/neotypes/generic/DeriveTupleMapSpec.scala
+++ b/tests/src/test/scala/neotypes/generic/DeriveTupleMapSpec.scala
@@ -4,6 +4,8 @@ import neotypes.BaseSynchronousSpec
 import neotypes.mappers.ResultMapper
 import neotypes.model.types._
 
+import scala.collection.immutable.SeqMap
+
 /** Base class for testing the derivation of a ResultMapper instance for tuples. */
 final class DeriveTupleMapSpec extends BaseSynchronousSpec {
 
@@ -13,7 +15,7 @@ final class DeriveTupleMapSpec extends BaseSynchronousSpec {
     val mapper = ResultMapper.tuple[Int, String, Boolean]
 
     val input = NeoMap(properties =
-      Map(
+      SeqMap(
         "int" -> Value.Integer(3),
         "str" -> Value.Str("Luis"),
         "bool" -> Value.Bool(true)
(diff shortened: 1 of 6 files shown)


tests/src/test/scala/neotypes/generic/DeriveTupleMapSpec.scala
```diff
@@ -0,0 +1,50 @@
...
+      )
+    )
+    val result = mapper.decode(input)
+
+    result.value shouldBe (Some(5), Right(true), List(0, 10), Map("foo" -> "bar", "baz" -> "quax"), None)
+  }
+
+  it should "not derive an instance for an untagged HList" in {
+    assertDoesNotCompile("import shapeless._; ResultMapper.tuple[Int :: String :: HNil]")
+  }
```
BalmungSan, line 49: This is unnecessary :)

BalmungSan (approved):
Thanks a lot!

tests/src/test/scala/neotypes/generic/DeriveTupleMapSpec.scala
turb: Fixup 427ad34dcb334d07ea0bc1ef86628d65e743b9d0

BalmungSan (approved):

